PR #3, opened 2026-06-26
docs: add agent guidance

jtenner:
## Summary

- Add compact `AGENTS.md` project guidance for low-footprint, performance-oriented WebAssembly JIT work.
- Add a commit skill focused on small atomic, measurable commits.
- Add agent workflow docs covering measurement, review, and documentation expectations.

## Tests

Not run; documentation-only change.

Signed-off-by: jtenner <3761339+jtenner@users.noreply.github.com>

Issue #1, opened 2026-06-25
perf: widen `v128` operations to best-available when possible


JairusSW:
Trigger on `0xFD` opcode and lower or widen vector operations to 256 or 512 bits

JairusSW:
solved with custom wasm instructions

PR #22, opened 2026-06-27
ir: add compact wasm SSA builder

jtenner:
## Summary

This branch introduces a compact high-level IR for WebAssembly functions under `src/core/compiler/ir`. The IR is intended as the next lowering boundary between the wasm decoder/validator and backend code generation: it keeps WebAssembly semantics explicit, represents operand-stack flow as SSA values with block parameters, preserves mutable locals as explicit effects for now, and records memory/table/global/call side effects so future optimization and codegen passes can be correct by construction.

It also includes the hardening fixes from the branch review:

- preserve stack height for unreachable untyped pops in the IR builder;
- bound `br_table` target allocation in validation and amd64 lowering;
- make the current single-table runtime support boundary explicit;
- preserve zero-length table metadata so `call_indirect` traps cleanly;
- reject active element expression segments instead of silently ignoring them;
- keep declared local runs compact in validation/backend metadata and bound backend local frames.

## Why add this IR?

The current amd64 compiler lowers directly from wasm bytecode. That is simple and fast to implement, but it makes later improvements difficult because the backend has to simultaneously:

1. parse wasm instructions;
2. model structured control flow;
3. preserve stack-polymorphic unreachable semantics;
4. track locals, globals, memory, tables, calls, traps, and host-call barriers;
5. choose native instructions/registers; and
6. maintain WasmWrapper/runtime ABI details.

The new IR splits those responsibilities. The IR builder is responsible for turning validated wasm bytecode into an auditable, verifier-checked graph. Later passes can consume a typed CFG instead of reparsing wasm. That gives us a safer place to implement optimizations and target-specific lowering without repeatedly re-deriving WebAssembly control-flow and stack rules.

The design goals match wago’s project priorities:

- **Correct wasm semantics first:** stack-polymorphism, block signatures, branch arities, memory/global/table/call effects, traps, and module-indexed references are all represented or rejected clearly.
- **Small predictable memory use:** variable-length operands live in shared pools; identifiers are `uint32`; declared locals stay run-length encoded; verifier dominance is linear-ish instead of bitset-per-block.
- **Fast hot paths later:** block parameters make stack values explicit at joins; effects give scheduling/codegen barriers; calls and memory/table operations are already annotated for backend lowering.
- **No cgo / simple runtime boundaries:** the IR models only the ABI-relevant facts needed by the existing Go+JIT runtime.
- **Auditable compiler changes:** `VerifyModule` and focused tests make IR invariants explicit before codegen depends on them.

### Module-level shape

`ir.Module` mirrors only metadata that later optimization/codegen needs:

- function types and per-function type indexes;
- imported function count;
- globals, memories, and tables;
- compact element/data metadata;
- local function IR bodies.

It is intentionally not a second full wasm AST. Large byte slices, names, and unsupported features are not copied into the IR unless later compiler stages need them.

### Function-level shape

Each `ir.Func` contains:

- `Blocks []Block`
- `Insts []Inst`
- `Values []Value`
- `ValueIDs []ValueID`
- `Edges []Edge`

Variable-length data is stored in shared contiguous pools and referenced by compact `Range { Start, Len }` values. For example:

- a block’s parameters are a range in `ValueIDs`;
- an instruction’s arguments/results are ranges in `ValueIDs`;
- a terminator’s branch edges are a range in `Edges`;
- each edge’s arguments are a range in `ValueIDs`.

This avoids per-instruction/per-edge slice headers and keeps retained IR memory proportional to actual content rather than allocation shape.

### IDs and values

The IR uses compact IDs:

- `ValueID uint32`
- `BlockID uint32`
- `InstID uint32`

A `Value` records:

- its wasm value type;
- whether it is defined by a block parameter, instruction result, or poison/unreachable placeholder;
- the defining block or instruction ID.

Poison values are used only to model stack-polymorphic unreachable code during construction. The verifier rejects reachable uses of poison values.

### Blocks and block parameters

WebAssembly structured control flow is lowered into a CFG where joins receive values via block parameters. This is the key SSA property of the IR.

For example, a wasm block/if/loop result becomes:

- a merge block with parameters for the result types;
- branch edges carrying the values into those parameters;
- later uses reading the merge block parameters.

This is a natural fit for Wasm because every structured label has a known branch signature. Instead of materializing a mutable operand stack at every join, the builder makes the transferred values explicit in the CFG.

Loops use their block parameter list for loop-carried values. A branch to a loop label targets the loop header with the loop’s input types. A branch to a non-loop label targets the merge block with the block’s result types.

### Locals intentionally remain explicit mutable state

The IR converts the operand stack to SSA-like values, but it does **not** yet convert wasm locals to SSA. `local.get`, `local.set`, and `local.tee` are represented as explicit instructions:

- `OpLocalGet` has `EffectReadLocal`;
- `OpLocalSet` has `EffectWriteLocal`;
- `OpLocalTee` has `EffectWriteLocal` and no read effect.

This is deliberate. WebAssembly locals are mutable state. Keeping them explicit at this first IR layer avoids an incomplete or surprising local-SSA transformation. Optimizers/codegen must treat local read/write effects as ordering constraints. A later local-SSA pass can remove local effects when it has full dominance/control-flow information.

Declared locals also remain compact: parameters are kept in the O(1) prefix, while declared locals retain their wasm run-length encoding. This prevents modules with huge local counts from allocating one `ValType` per local during IR construction.

### Terminators

Blocks end in one of:

- `TermBr`
- `TermCondBr`
- `TermSwitch`
- `TermReturn`
- `TermTrap`

Branch-like terminators reference `Edge` records. Each edge has a target block and a range of argument values for the target block parameters.

Branches to the function label are either direct returns or, for branch-like terminators that require block targets (e.g. `br_if`/`br_table` cases), go through a synthetic return block. The builder reuses a single synthetic return block per function and marks it with `BlockSyntheticReturn`, making it easy for later CFG cleanup/codegen to identify.

### Instructions and effects

`ir.Op` covers the currently supported lowering surface:

- constants;
- integer/float unary, binary, comparisons, tests;
- conversions/reinterpretations;
- select;
- loads/stores;
- memory size/grow/copy/fill;
- globals;
- direct/imported/indirect calls;
- locals.

Auxiliary opcode details are packed into `Aux`/`Aux2` to keep `Inst` compact. Examples:

- numeric sub-op and value type are packed together;
- memory kind, alignment, memory index, and offset are packed together;
- indirect-call type/table indexes are packed together.

Effects are explicit flags:

- `EffectCanTrap`
- `EffectReadMem` / `EffectWriteMem`
- `EffectReadGlobal` / `EffectWriteGlobal`
- `EffectReadTable` / `EffectWriteTable`
- `EffectCall`
- `EffectHost`
- `EffectReadLocal` / `EffectWriteLocal`

These flags are important because future passes should be able to reorder or fold pure computations without crossing traps, host calls, memory state, globals, table reads, or local state. For example:

- `memory.size` is modeled as `EffectReadMem` because a preceding `memory.grow` changes its result.
- `memory.grow` is read+write+trap.
- imported calls have `EffectHost` and `EffectCall`.
- indirect calls have `EffectReadTable` and `EffectCanTrap`.
- integer division/rem can trap.

### Verifier

The verifier checks both shape and semantics:

- valid entry block;
- local layout and declared local runs;
- all value ranges and edge ranges are in bounds, including `uint32` wrap protection;
- instruction arity/type rules;
- opcode auxiliary ranges;
- required effects for memory, globals, locals, calls, host calls, and indirect calls;
- module-indexed references when module metadata is available;
- terminator arity and target parameter compatibility;
- no poison values in reachable uses;
- each instruction belongs to exactly one block;
- dominance of every value use.

Dominance is computed with immediate dominators instead of a full bitset per block. This keeps verifier memory linear in blocks/edges and avoids fixed-point bitset allocations for branch-heavy modules.

## How the builder lowers wasm into this IR

At a high level, `BuildModule` copies compact module metadata and calls `buildFunc` for each local function.

For each function:

1. Create a parameterless entry block. Function parameters are locals at this IR layer, not SSA entry parameters.
2. Initialize the label stack with the function label and result types.
3. Parse wasm bytecode sequentially with a small operand stack of `ValueID`s.
4. For each expression/control construct, build blocks, block parameters, instructions, and terminators.
5. On fallthrough or branch, pop values matching the label/block signature and attach them to edge arguments.
6. On unreachable code, use poison values for stack-polymorphic pops/results without allowing them to escape reachable IR.
7. Terminate any unreachable/dead blocks with traps so every block has a terminator.
8. Trim retained slice slack for memory-sensitive deployments.

Some examples:

### Straight-line arithmetic

Wasm:

```wat
local.get 0
i32.const 1
i32.add
```

IR shape:

```text
%0:i32 = local.get 0        ; EffectReadLocal
%1:i32 = const i32 1
%2:i32 = ibinary.add %0, %1
```

The operand stack has become explicit SSA value IDs.

### If/else with result

Wasm block result values become merge-block parameters:

```text
condbr %cond, b_then(args...), b_else(args...)

b_then(...):
  ...
  br b_merge(%then_result)

b_else(...):
  ...
  br b_merge(%else_result)

b_merge(%result):
  ... use %result ...
```

The backend no longer has to rediscover the phi-like value at the join. It is the merge block parameter.

### br_table

`br_table` becomes `TermSwitch` with one edge for each label plus the default edge. The target vector is bounded before allocation because each target must consume at least one byte of LEB128 encoding.

### Unreachable code

When a point is unreachable, wasm stack-polymorphism allows instructions to pop arbitrary typed values. The builder returns poison placeholders at the current control-frame height rather than mutating enclosing stack state. The verifier rejects poison if it appears in reachable instruction/edge/return uses.

## How this would lower to machine instructions

This branch adds the IR and verifier; it does not yet replace the existing direct wasm-to-amd64 backend. The intended lowering pipeline is:

```text
wasm module
  -> decode/validate
  -> ir.BuildModule
  -> ir.VerifyModule
  -> optional IR cleanup/optimization passes
  -> target-specific instruction selection / register allocation
  -> runtime ABI wrapper emission
```

### Mapping IR values to machine locations

A backend can assign each `ValueID` to one of:

- a machine register;
- an XMM register for float values;
- a stack spill slot;
- a constant/immediate rematerialization;
- a known local/global/memory address when safe.

Instruction results define new values. Instruction arguments reference existing values. Because dominance is verified, a backend can assume each use has a valid reaching definition.

### Lowering block parameters

Block parameters are the IR equivalent of phi nodes. There are two straightforward lowering strategies:

1. **Edge move lowering:** assign each block parameter a canonical location at the start of the target block. Every predecessor edge emits moves from edge argument locations into those canonical parameter locations before branching.
2. **SSA destruction / parallel copy pass:** compute block parameter copies as parallel moves, resolve cycles with a scratch register/spill, then emit target-specific moves.

This is similar to what the current direct amd64 backend already does with canonical operand stack slots at structured joins, but the IR makes the data flow explicit and no longer tied to reparsing structured wasm.

For example:

```text
br b3(%7, %8)
```

would lower on that edge to something like:

```text
move location(%7) -> location(b3.param0)
move location(%8) -> location(b3.param1)
jmp b3
```

For `condbr`, the backend emits the conditional branch and places the edge copies on each outgoing path. For `switch`, each case path gets copies into its target parameter locations before the jump.

### Lowering locals

Since locals remain explicit mutable state in this IR layer, an initial backend can lower them much like the current backend:

- allocate local slots in the native frame or a separate local area;
- copy function parameters into local slots on entry;
- zero declared locals;
- lower `OpLocalGet` to a load from the local slot;
- lower `OpLocalSet` to a store to the local slot;
- lower `OpLocalTee` to a store plus forwarding the same value as the instruction result.

Because local read/write effects are explicit, a scheduler/optimizer must not move local operations across each other in ways that change mutable local semantics. A later local-SSA pass can transform local state into SSA values where profitable, then remove some local effects.

### Lowering numeric operations

Numeric instructions are direct instruction-selection candidates:

- `OpIUnary` -> `clz/ctz/popcnt` sequences or extension operations;
- `OpIBinary` -> integer ALU operations, shifts, rotates, division/rem with trap paths;
- `OpICmp` / `OpITest` -> compare/test + setcc or branch fusion;
- `OpFUnary` / `OpFBinary` / `OpFCmp` -> SSE/AVX scalar operations;
- `OpConvert` / `OpReinterpret` -> target conversion instructions or helper sequences.

The packed aux kind tells the backend exactly which wasm operation is being lowered, and the value type tells whether it is i32/i64/f32/f64.

### Lowering memory operations

Memory ops carry:

- memory operation kind;
- alignment;
- memory index (currently single-memory only);
- static offset;
- address/value arguments;
- trap/read/write effects.

A backend would lower `OpLoad`/`OpStore` as:

1. materialize address value;
2. add static offset with 64-bit arithmetic;
3. bounds check against linear-memory size;
4. trap on out-of-bounds;
5. emit the load/store with extension/truncation according to `MemOp`.

Bulk memory operations (`memory.copy`, `memory.fill`) already have explicit read/write/trap effects and i32 operands in the IR, making them natural calls or inline loops/`rep movsb`/`rep stosb` sequences.

### Lowering globals

Globals lower through the runtime global pointer table:

- `OpGlobalGet` loads the global cell pointer and then loads the value bits according to the global type.
- `OpGlobalSet` loads the cell pointer and stores the value bits.

The verifier checks global index/type validity when module metadata is available and enforces read/write effects.

### Lowering direct, imported, and indirect calls

Calls are split by opcode:

- `OpCall` for local/direct wasm calls;
- `OpCallImport` for host/import calls;
- `OpCallIndirect` for table dispatch.

A backend can use this split to lower each call boundary correctly:

- `OpCall`: marshal arguments/results to the internal calling convention and patch direct call relocations after module layout.
- `OpCallImport`: marshal arguments/results through the host-call ABI and keep `EffectHost` as a scheduling barrier.
- `OpCallIndirect`: load table descriptor, bounds-check index, check signature ID, check null code pointer, then call through the loaded code pointer.

The IR records indirect-call type/table metadata and canonical type ID in aux fields, and the verifier requires `EffectReadTable`, `EffectCall`, and `EffectCanTrap`.

### Trap lowering

Any instruction with `EffectCanTrap` can lower to a fast path plus a trap exit path. `TermTrap` lowers to the runtime trap sequence directly. This keeps trap-capable operations visible to optimization passes: they are not pure, and reordering them across other observable effects must be justified.

### Why this helps future optimization

Once wasm is in this IR, future passes can be written against explicit data/control/effect information:

- constant folding for pure numeric ops;
- local-SSA conversion;
- dead-code elimination that respects effects;
- block merging and synthetic return cleanup;
- branch simplification;
- memory bounds-check folding where safe;
- better register allocation from SSA values instead of a simulated wasm stack.

The verifier is the safety net: passes can mutate IR and re-run verification before machine code emission.

## Important support boundaries in this branch

This branch keeps unsupported behavior explicit rather than silently miscompiling:

- Multi-memory remains rejected.
- Imported/multiple tables are rejected until the runtime has table import/multi-table descriptors.
- Active element expression segments are rejected until they are lowered into table entries.
- Non-zero `call_indirect` table indexes are rejected in direct amd64 lowering as a defensive boundary.
- Backend local frames are bounded (`maxCompiledLocals`) while validation/IR construction remain compact for huge local declarations.

## Testing

Ran:

```sh
go test ./...
```

Additional targeted coverage includes:

- IR build/verify golden and negative tests;
- dominance verification tests;
- instruction/effect verification tests;
- wrapped range rejection tests;
- br_table allocation-bound tests;
- unreachable stack-polymorphism regression tests;
- table support-boundary tests;
- zero-length table `call_indirect` trap test;
- active element expression rejection test;
- compact huge-local validation/backend rejection tests.

## Notes for reviewers

The new IR package is intentionally conservative. It does not try to solve every compiler problem in this branch. In particular, locals are not SSA yet, and the existing amd64 backend is not replaced here. The main review questions are:

1. Are the IR invariants explicit enough for future passes/codegen?
2. Are effects conservative enough to prevent invalid reordering?
3. Are unsupported wasm features rejected at clear boundaries?
4. Are the retained-memory choices appropriate for small devices?
5. Is the verifier strong enough to be trusted between optimization and machine-code lowering?


jtenner:
@codex can you please do a deep review of this pr?

src/core/compiler/backend/amd64/fp.go
```diff
@@ -290,7 +290,17 @@ func (g *cg) fstore(r *wasm.Reader, f64 bool) error {
...
+		return isFloatType(g.localParams[i])
+	}
+	rem := uint64(i - len(g.localParams))
+	for _, run := range g.localRuns {
+		if rem < uint64(run.Count) {
+			return isFloatType(run.Type)
+		}
+		rem -= uint64(run.Count)
+	}
+	return false
```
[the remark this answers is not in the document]
jtenner: Addressed in 236ad1b. The compact local lookup now returns `(type, ok)` and checks `i < 0 || i >= nLocals`; `local.get`, `local.set`, and `local.tee` now return `amd64: unknown local N` before computing frame offsets. Added `TestCompileRejectsOutOfRangeCompactLocalLookup`.

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1737 @@
...
+	b := &Builder{m: m, out: out}
+	return b.buildFunc(uint32(localFuncIdx))
+}
+
+func (b *Builder) buildFunc(localIdx uint32) (*Func, error) {
+	if int(localIdx) >= len(b.m.FuncTypes) || int(localIdx) >= len(b.m.Code) {
+		return nil, fmt.Errorf("ir: local function index %d out of range", localIdx)
+	}
+	typeIdx := b.m.FuncTypes[localIdx].Index
+	ftc, ok := b.m.LocalFuncType(int(localIdx))
```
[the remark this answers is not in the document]
jtenner: Addressed in 236ad1b. The wasm type helpers now resolve TypeIdx through flattened recursive-group subtypes, CanonicalTypeID iterates flattened indexes, and IR module metadata preserves flattened subtype positions (including non-function placeholders). Added `TestBuildFuncTypeUsesFlattenedRecGroupSubtypeIndex`.

jtenner:
@codex can you do another review?

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1731 @@
...
+	for i := range m.Imports {
+		if m.Imports[i].Type.Kind == wasm.ExternMem {
+			n++
+		}
+	}
+	return n
+}
+
+func (b *Builder) readZeroMemoryImmediate() (uint32, error) {
+	mem, err := b.r.Byte()
```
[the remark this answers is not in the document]
jtenner: Fixed in the current branch. `readZeroMemoryImmediate` now decodes the memory immediate with `b.r.U32()` instead of reading one byte, so multi-byte ULEB encodings of memory index 0 (for example `0x80 0x00`) are accepted, while non-zero indexes are still rejected at the IR boundary. Covered by `TestBuildMemoryImmediatesUseULEBEncoding`.

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1731 @@
...
+		if err != nil {
+			return err
+		}
+		if _, err := b.memoryType(mem); err != nil {
+			return err
+		}
+		if b.reachable {
+			// memory.size observes mutable memory state: a preceding memory.grow can
+			// change the result, so it must not be modeled as a pure instruction.
+			b.pushValues(b.addInst(OpMemorySize, uint64(mem), 0, nil, []wasm.ValType{wasm.I32}, EffectReadMem))
```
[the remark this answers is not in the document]
jtenner: Fixed in the current branch. The builder now derives memory operand/result widths from the module memory type via `memoryAddrType(mt)` / `wasm.MemoryAddrType`: `memory.size` and `memory.grow` produce `i64` for memory64, loads/stores pop `i64` addresses for memory64, and bulk memory uses the correct source/destination/length address types. The verifier mirrors this with `verifyMemoryAddrType`. Covered by `TestBuildMemory64UsesI64AddressesAndSizes`.

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1731 @@
...
+		ref := tableRefType(tt)
+		if wasm.RefVal(ref) != wasm.FuncRef {
+			return fmt.Errorf("call_indirect table %d has element type %s", tbl, wasm.RefVal(ref))
+		}
+		ftc, ok := b.m.TypeFunc(ti)
+		if !ok {
+			return fmt.Errorf("unknown type %d", ti)
+		}
+		ft := funcTypeFromComp(ftc)
+		callee, err := b.popTyped(wasm.I32)
```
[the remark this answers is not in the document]
jtenner: Fixed in the current branch. `call_indirect` now pops the callee/table index using `tableAddrType(tt)`, which maps table64 tables to `i64` and table32 tables to `i32`. The verifier also checks the final IR argument against `tableAddrType(m.Tables[tableIdx])`, so builder and verifier agree on the downstream codegen contract. Covered by the `table64 index` subtest in `TestBuildCallIndirectReferenceAndAddressTypes`.

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1731 @@
...
+		if err != nil {
+			return err
+		}
+		tt, err := b.tableType(tbl)
+		if err != nil {
+			return err
+		}
+		ref := tableRefType(tt)
+		if wasm.RefVal(ref) != wasm.FuncRef {
+			return fmt.Errorf("call_indirect table %d has element type %s", tbl, wasm.RefVal(ref))
```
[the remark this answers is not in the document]
jtenner: Fixed in the current branch. The builder no longer compares `wasm.RefVal(ref)` to the bare `funcref` alias. It now uses `isFuncRefTableType(b.m, ref)`, which accepts absolute `func`/`nofunc`, module type indexes that resolve to function types, and function def-types. The verifier has the matching IR-module-aware check `irIsFuncRefTableType`. Covered by the `non-bare funcref table` subtest in `TestBuildCallIndirectReferenceAndAddressTypes` and the new indexed-heap verifier tests.

JairusSW (approved):

src/core/compiler/ir/verify.go
```diff
@@ -0,0 +1,1292 @@
...
+		}
+	}
+	return idx
+}
+
+func verifyFunc(f *Func, m *Module) error {
+	if f == nil {
+		return fmt.Errorf("ir: nil func")
+	}
+	if int(f.Entry) >= len(f.Blocks) {
```
JairusSW, line 150: **Soundness gap (verifier wrongly accepts):** the entry block's `Params` are never validated to be empty (nor matched against `f.Sig`). Block params are only constrained by the *edge* arity/type checks in `verifyEdges`, and the entry block has no incoming edges — so a `Func` whose entry block carries a `Params` range passes every check, and since entry dominates everything, an entry param flowing into any use is accepted. Codegen trusting verified IR would materialize an undefined value.

Confirmed with a minimal repro: a `Func` with an undefined i32 entry-block param flowing into `return` passes `VerifyFunc` with `nil`. The builder always makes entry parameterless, so this only bites hand-built/optimizer-mutated IR — which is exactly what this gate exists to defend against.

Suggest rejecting a non-empty `Params` range on `f.Entry` here (or validating it against `f.Sig.Params` if entry params become the ABI).

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1824 @@
...
+	return b.scratchValues, nil
+}
+func (b *Builder) popTyped(t wasm.ValType) (ValueID, error) {
+	if !supportsIRValType(t) {
+		return InvalidValue, fmt.Errorf("unsupported IR value type %s", t)
+	}
+	if !b.reachable && len(b.stack) == b.ctrlH[len(b.ctrlH)-1] {
+		return b.newValue(t, ValueDefPoison, 0), nil
+	}
+	if len(b.stack) == 0 {
```
JairusSW: **Reachable pops aren't bounded by the control-frame height.** The `ctrlH` floor is only consulted on the unreachable branch (line 1495); in reachable code the sole underflow guard is `len(b.stack) == 0`. A crafted module — e.g. an empty-typed `block` containing `i32.const 0; i32.add` — can therefore pop an operand belonging to an *enclosing* frame: `add` pops the const, then pops the enclosing block's value (reachable, stack non-empty → allowed), and the block closes balanced, so no error is returned and the IR silently miscompiles.

Not reachable through the normal validate→build pipeline, but the surrounding comments claim this builder is "a defensive lowering boundary when called without a prior validator pass," and every other malformed-input class here is rejected. Suggest adding a `len(b.stack) <= b.ctrlH[len(b.ctrlH)-1]` rejection in both `popTyped` and `popMaybe`. Same applies to `popMaybe` below (line ~1519).

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1824 @@
...
+		ft, err := b.funcType(fi)
+		if err != nil {
+			return err
+		}
+		args, err := b.popValues(ft.Params)
+		if err != nil {
+			return err
+		}
+		if b.reachable {
+			res := b.addInst(callOp(fi, uint32(b.m.ImportedFuncCount())), uint64(fi), 0, args, ft.Results, EffectCanTrap|EffectCall|hostEffect(fi, uint32(b.m.ImportedFuncCount())))
```
JairusSW, line 721: **Call ops carry no memory/global effects** — correctness depends on an unstated contract. `OpCall`/`OpCallImport` (here) and `OpCallIndirect` (line 761) set `EffectCall` but not `EffectReadMem/WriteMem/ReadGlobal/WriteGlobal`, even though calls touch all of those. This is only sound if *every* future consumer treats `EffectCall` as a maximal barrier subsuming all memory/global/table effects. A disambiguation/scheduling pass keying off the specific flags ("does this inst write memory?") could legally hoist a load/store or global access across a call.

Please either document on `EffectCall` (in op.go) that it subsumes all mem/global/table effects, or set the flags explicitly. Worth nailing down before any optimization pass is wired in.

src/core/compiler/wasm/types.go
```diff
@@ -273,11 +273,13 @@ type Limits struct {
 	Min    uint64
 	Max    *uint64
 	Addr64 bool
+	HasMax bool // legacy test/helper compatibility; Max remains authoritative when set.
 }
 
 type TagType struct{ Type TypeIdx }
 type GlobalType struct {
 	Type    ValType
+	Val     ValType // legacy test/helper compatibility; Type is authoritative when set.
```
JairusSW: **Test-only fields on a production struct.** `GlobalType.Val` (here), `TableType.Elem` (line 291), and `Limits.HasMax` (line 276) are written only by tests — the decoder always populates the canonical `.Type`/`.Ref`, so the fallback branches in `GlobalValueType`/`TableRefType` are dead in production. This adds permanent ambiguity to core decoder types purely for IR-test construction ergonomics, which cuts against the AGENTS.md "be strict, don't soften for tests" stance. Prefer constructing the canonical fields in the IR tests and dropping these three fields + their fallbacks.

src/core/compiler/ir/verify.go
```diff
@@ -0,0 +1,1292 @@
...
+		}
+		for i := 0; i < a; i++ {
+			if argt(i) != rest(0) {
+				return fmt.Errorf("inst %d type mismatch", id)
+			}
+		}
+		return nil
+	}
+	switch in.Op {
+	case OpConst:
```
JairusSW, line 384: **Aux canonicalization is inconsistent.** `memory.size/grow/fill` reject stray high bits (`uint64(uint32(in.Aux)) != in.Aux`), but `OpConst` here (and `OpCall`/arith/convert/select) don't constrain the unused `Aux` bits at all. Sound only as long as codegen reads via the same low-bit accessors; for a trust gate, assert the unused bits are zero here too so the metadata is provably canonical.

JairusSW (commented):
## Review: SSA IR tier

Reviewed architecture + correctness across `ir/build.go`, `ir/verify.go`, the front-end refactors, and the single-pass backend hardening. Builds clean, `vet`/`gofmt` clean, full suite passes, IR package at 86.9% coverage.

**Fit with the single-pass (Valent-Block) path is good:** the IR is cleanly isolated (nothing imports it yet), and the front-end type-ID refactor (`typeFunc` flattening, `CanonicalTypeID`) is *shared* by both tiers rather than forked, so a future IR backend can't disagree with the single-pass backend on `call_indirect` signature identity. Metadata (`CanonicalTypeIDs`, imports-first index spaces, table/elem/data descriptors) mirrors what `Instantiate` already needs. Locals-as-explicit-stateful-ops and the effect-flag model are sound foundations for an optimizing tier.

Design quality is high — allocation-conscious shared pools, capped+trimmed capacities, run-length locals, cached return block. The unreachable/stack-polymorphic handling is carefully done.

### Substantive (address before this becomes load-bearing)
- **Verifier accepts an uninitialized entry-block param** (inline on `verify.go`). Confirmed with a minimal repro: a `Func` whose entry block has an undefined param flowing into `return` passes `VerifyFunc` with `nil`. Soundness hole in the trust gate.
- **Builder reachable pops aren't bounded by the control-frame height** (inline on `build.go`). Defeats the documented "defensive without a validator pass" claim; not reachable via the normal validate→build pipeline.
- **Call effect-flags rely on an undocumented "EffectCall is a maximal barrier" contract** (inline on `build.go`). Pin this down before wiring any optimization pass.

### Cleanup / process
- **Test-only legacy fields on production structs** (inline on `wasm/types.go`): `GlobalType.Val`, `TableType.Elem`, `Limits.HasMax` are written only by tests; the decoder always sets the canonical fields, so the fallback branches are dead in production. Prefer constructing canonical fields in tests and dropping these.
- **Docs not updated**: `ARCHITECTURE.md:110` still says *"There is no separate IR"*, and nothing documents the new tier's purpose/scalar-only scope/intended consumer. AGENTS.md asks for doc updates on architecture changes.
- **Aux canonicalization is inconsistent** in the verifier (inline) — `memory.size/grow/fill` reject stray high bits; `OpConst`/`OpCall`/arith don't.

### Nits
- `EffectCanTrap` on `OpMemoryGrow` (inline) — grow returns -1, doesn't trap.
- `popValues` returns the shared `scratchValues` buffer — safe today, latent aliasing foot-gun.
- `lowerBrIf` rebuilds the stack with `append(b.stack, …)` instead of `append(b.stack[:height], …)` — correct but asymmetric vs the other lowerings.
- The `rep stosb` local-zeroing rewrite is correct (covers exactly the declared-local region), but has startup overhead vs inline `mov`s for the common 1–3-local case — consider a small-count threshold.

Full verdict: strong, careful work; #1 (verifier hole) and #3 (call effect contract) are the two to resolve first.

src/core/compiler/ir/build.go
```diff
@@ -0,0 +1,1824 @@
...
+		if err != nil {
+			return err
+		}
+		addr := memoryAddrType(mt)
+		pages, err := b.popTyped(addr)
+		if err != nil {
+			return err
+		}
+		if b.reachable {
+			b.pushValues(b.addInst(OpMemoryGrow, uint64(mem), 0, []ValueID{pages}, []wasm.ValType{addr}, EffectCanTrap|EffectReadMem|EffectWriteMem))
```
JairusSW: Nit: `memory.grow` doesn't trap — it returns -1 on failure — so `EffectCanTrap` is an over-approximation that can needlessly inhibit later code motion. The `EffectReadMem|EffectWriteMem` pair is correct (it must be a full memory barrier).

jtenner:
Addressed Jairus's review feedback in commit eb7ee17 (`ir: address PR verifier feedback`). Per-comment responses:

- Verifier accepts uninitialized entry-block param: fixed by validating any entry-block params against `Func.Sig.Params`, with regression coverage for mismatched/unexpected entry params.
- Reachable pops not bounded by control-frame height: fixed `popTyped`/`popMaybe` to reject reachable pops at/below the current control-frame floor, with a malformed-bytecode regression test.
- Call effect flags contract: documented `EffectCall` in `op.go` as a maximal ordering barrier that subsumes memory/global/table/local/host-visible state.
- Test-only legacy wasm fields: removed `GlobalType.Val`, `TableType.Elem`, and `Limits.HasMax`; updated IR/wasm tests to construct canonical `Type`/`Ref`/`Max` metadata.
- Aux canonicalization: added verifier checks for scalar aux high bits where bits are unused (`const` i32/f32, local/global/call indexes, kind+type aux, select, call_indirect aux2).
- `memory.grow` trap effect: removed `EffectCanTrap` from builder/verifier expectations and tests; it remains a read/write memory barrier.
- `popValues` scratch aliasing: changed `popValues` to return an owned slice and removed the shared scratch field.
- `lowerBrIf` stack rebuild nit: made the false-continuation stack rebuild use an explicit saved height (`append(b.stack[:height], ...)`) like the other lowerings.
- `rep stosb` small-local threshold: added a small-count inline zero-store path for 1–3 declared locals, keeping `rep stosb` for larger local ranges.
- Docs: updated `ARCHITECTURE.md` to describe the new scalar SSA IR tier, scope, verifier role, and current isolation from the production hot path.

Proof: `go test ./...`

src/core/compiler/ir/verify.go
jtenner: Fixed in eb7ee17: the verifier now validates entry-block params against `Func.Sig.Params` when present, so hand-built/optimizer-mutated IR cannot introduce an undefined entry value. Added regression coverage for unexpected/mismatched entry params.

src/core/compiler/ir/build.go
jtenner: Fixed in eb7ee17: `popTyped` and `popMaybe` now reject reachable pops at/below the current `ctrlH` floor, while preserving stack-polymorphic poison behavior for unreachable code. Added the crafted malformed-block regression test.

src/core/compiler/ir/build.go
jtenner: Fixed in eb7ee17 by documenting the contract on `EffectCall` in `op.go`: it is a maximal ordering barrier that subsumes memory/global/table/local/host-visible state for future consumers.

src/core/compiler/wasm/types.go
jtenner: Fixed in eb7ee17: removed `GlobalType.Val`, `TableType.Elem`, and `Limits.HasMax`, removed the fallback helper logic, and updated tests to build canonical `Type`/`Ref` metadata directly.

src/core/compiler/ir/verify.go
jtenner: Fixed in eb7ee17: verifier now rejects non-canonical unused aux bits for scalar metadata (i32/f32 consts, local/global/call indexes, kind+type aux, select, and call_indirect aux2).

src/core/compiler/ir/build.go
jtenner: Fixed in eb7ee17: `memory.grow` now carries `EffectReadMem|EffectWriteMem` without `EffectCanTrap`; verifier/tests were updated to match.

Issue #2, opened 2026-06-25
perf: simd-backed memory operations


JairusSW:
`memory.copy` and `memory.fill` can really benefit from simd

PR #25, opened 2026-06-29
perf(amd64): pin integer locals in registers (Valent-Block)

JairusSW:
## What

Ports WARP's **locals-in-registers** optimization into the wago single-pass backend. The first few integer locals (params first — the hottest) are **pinned to dedicated registers** (`RBX`/`R13`/`R14`/`R15`) instead of living in frame slots, so `local.get`/`local.set`/`local.tee` avoid memory traffic and operands fold the register directly (e.g. `add dst, R_x`).

Until now every local lived in a frame slot (`vLocal`), read/written on each access. This is the flagship perf item from the porting backlog.

## Design (faithful in spirit; v1 scoped for low risk)

- **New lazy operand kind `vPinned`**: `local.get` of a pinned local pushes a reference to its register (emits nothing); source consumers fold it as an RHS, and it is *copied* (never reused as a mutable dest) when materialized.
- Pinned registers are removed from the scratch pool (`cg.reserved`), so they're never handed out as temporaries.
- **Canonical-slot joins need no extra work**: a pinned local always lives in the same register on every edge into a join, so branches/loops stay consistent for free. `flush()` copies the register to the operand slot but leaves the local in place.
- **Calls** clobber the registers (they're plain scratch in the callee), so `emitWrapperCall` spills pinned locals to their frame slots around the call and reloads after. Host calls (deferred log) don't clobber, so they're untouched.
- `materializeLocalRefs` captures pending `vPinned` reads before `local.set` overwrites the register (the lazy-aliasing hazard).
- Float locals stay frame-resident (integer-only in v1).

## Tests

- New `pinned_test.go`: >pool-size locals boundary (pinned + frame-resident mixed in one expr), double-use of one pinned local, alias capture, loop back-edge consistency, if/else join.
- Existing recursion / AssemblyScript payloads exercise pinned params across calls.
- Full `go test ./...` green incl. `-race`; CLI spot-checks `fib(25)=75025`, `sumto(100)=5050`.

## Performance (bench/ Exec suite, count=8, benchstat) — geomean **−17.1%** sec/op

| benchmark | before → after | delta |
|---|---|---|
| Exec/arith.run | 4.26µs → 1.48µs | **−65.3%** |
| Exec/fib_iter.fib | 65.8ns → 35.6ns | **−45.8%** |
| Exec/memory.sum | 821ns → 462ns | **−43.8%** |
| ExecFibLoop_wago | 34.7ns → 21.4ns | **−38.4%** (wazero 43.7ns) |
| Exec/globals.accumulate | 2.48µs → 1.76µs | **−29.1%** |
| Exec/tiny.add | 26.7ns → 25.1ns | **−6.1%** |

Regressions are confined to **call-heavy** paths — the known cost of spilling pinned locals around calls (option A): `ExecFibRec_wago` **+3.6%**, `Exec/dispatch.apply` (call_indirect) **+7.5%**. A register-based internal call ABI (a later port in this effort) is the natural fix. Exec allocs unchanged (0).

---
Port #2 of the WARP singlepass porting effort (interleaved feature/perf, one PR per port). Independent of #24 (sign-extension ops).

### DIFF
--- a/src/core/compiler/backend/amd64/compile.go
+++ b/src/core/compiler/backend/amd64/compile.go
@@ -25,11 +25,22 @@ var scratch = []Reg{RAX, RCX, RDX, RBX, R8, R9, R10, R11, R13, R14, R15}
 type vkind uint8
 
 const (
-	vConst vkind = iota // immediate constant (not yet materialized)
-	vLocal              // reference to a local's frame slot (lazy)
-	vReg                // value resident in a scratch register
-	vSpill              // value materialized in its canonical frame slot
+	vConst  vkind = iota // immediate constant (not yet materialized)
+	vLocal               // reference to a local's frame slot (lazy)
+	vReg                 // value resident in a scratch register
+	vSpill               // value materialized in its canonical frame slot
+	vPinned              // reference to a local pinned in a reserved register (lazy)
 )
+
+// regNone marks a local that is not pinned to a register (lives in its frame slot).
+const regNone Reg = 0xFF
+
+// pinnedLocal records an integer local kept resident in a dedicated register
+// for the whole function (spilled to its frame slot only around calls).
+type pinnedLocal struct {
+	local int
+	reg   Reg
+}
 
 type ventry struct {
 	kind  vkind
@@ -55,6 +66,9 @@ type cg struct {
 	nResults    int
 	relocs      []callReloc // internal call sites to patch at module layout
 	localFloat  []bool      // per-local: f32/f64?
+	localReg    []Reg       // per-local: pinned register, or regNone if frame-resident
+	pinned      []pinnedLocal
+	reserved    [16]bool // GPRs dedicated to pinned locals (not allocatable as scratch)
 }
 
 // Frame layout: saved ABI pointers, locals, then operand-stack spill slots.
@@ -73,7 +87,7 @@ func (g *cg) freeReg(r Reg) { g.busy[r] = false }
 
 func (g *cg) allocRegExcept(except Reg) Reg {
 	for _, r := range scratch {
-		if r != except && !g.busy[r] {
+		if r != except && !g.busy[r] && !g.reserved[r] {
 			g.busy[r] = true
 			return r
 		}
@@ -123,17 +137,27 @@ func (g *cg) loadInto(dst Reg, e ventry) {
 			g.a.MovReg64(dst, e.reg)
 			g.freeReg(e.reg)
 		}
+	case vPinned:
+		if e.reg != dst { // pinned reg is shared storage; copy out, never free
+			g.a.MovReg64(dst, e.reg)
+		}
 	case vSpill:
 		g.a.Load64(dst, RBP, g.slotOff(e.slot))
 	}
 }
 
 // materializeLocalRefs prevents lazy local.get entries from seeing later writes.
+// It captures pending reads of local x — both frame-resident (vLocal) and
+// register-pinned (vPinned) — to their canonical slots before x is overwritten.
 func (g *cg) materializeLocalRefs(x int) {
 	for i := range g.st {
-		if g.st[i].kind == vLocal && g.st[i].local == x {
+		switch {
+		case g.st[i].kind == vLocal && g.st[i].local == x:
 			g.a.Load64(RSI, RBP, g.localOff(x))
 			g.a.Store64(RBP, g.slotOff(i), RSI)
+			g.st[i] = ventry{kind: vSpill, slot: i}
+		case g.st[i].kind == vPinned && g.st[i].local == x:
+			g.a.Store64(RBP, g.slotOff(i), g.st[i].reg)
 			g.st[i] = ventry{kind: vSpill, slot: i}
 		}
 	}
@@ -264,6 +288,8 @@ func (g *cg) applyALU(d aluDesc, dst Reg, src ventry, w bool) {
 	case vReg:
 		g.a.AluRR(d.rr, dst, src.reg, w)
 		g.freeReg(src.reg)
+	case vPinned:
+		g.a.AluRR(d.rr, dst, src.reg, w) // pinned reg as RHS; shared storage, never freed
 	case vLocal:
 		g.a.AluRM(d.rm, dst, RBP, g.localOff(src.local), w)
 	case vSpill:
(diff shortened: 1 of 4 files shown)


JairusSW:
@jtenner it would be interesting to actually measure which locals are hot instead of just doing the first 4 or so. i was thinking that'd tie nicely into a higher, more intelligent tier

jtenner:
## Overall read

I like the direction. The implementation has the right shape for this optimization: pinned locals are represented as a distinct lazy operand (`vPinned`), the allocator excludes their registers, materialization copies rather than consumes them, control-flow flushes copy pinned values into canonical operand slots, and internal wasm calls spill/reload pinned locals around the callee. Those are the invariants I would expect for a low-risk first version.

I would **not block on the implementation as-is**, but I would ask for a couple of targeted tests before merging. The main risk is not straight-line local access; the main risk is preserving caller-pinned local state across generated-to-generated wasm calls, especially when the callee uses all the same pinned registers. The PR does implement the spill/reload in `emitWrapperCall`, but the new `pinned_test.go` tests mostly exercise `CompileFunction`, which cannot cover internal-call relocations.

CI is green for build/test/lint, and the CI card reports 597 tests passing with no failures, but the CI report does not include benchmark data, so I would treat the performance table in the PR body as author-supplied local benchstat rather than a verified CI artifact.

## Things that look correct

The lazy aliasing hazard is handled in the right place. `materializeLocalRefs` now captures pending `vPinned` reads into operand-stack spill slots before `local.set` or `local.tee` overwrites the pinned register, which is exactly the failure mode this change introduces.

The register allocator change is also the right minimal mechanism. Keeping `RBX/R13/R14/R15` in the `scratch` list but marking them reserved means the rest of the allocator logic still works, while `allocRegExcept` will not hand pinned registers out as temporaries.

The Go/native boundary appears safe in the current runtime model. Generated functions themselves only save `RBP`, but `enterNative` saves/restores `BX`, `R13`, `R14`, and `R15` around the whole foreign-stack call, so using those registers inside native wasm code is fine as long as all entry goes through `Engine.Call`.

## Main correctness gap: internal calls need a direct regression test

The code comments say generated callees clobber the caller’s pinned-local registers, and `emitWrapperCall` stores all pinned locals to their frame slots before the call and reloads them after. That is the highest-value invariant to test directly.

I would add a module-level helper because the current `runI32` helper uses `CompileFunction`, which rejects internal calls. `CompileModule` returns one code blob plus per-function entry offsets, and `runtime.MapCode` returns the base pointer, so the helper can enter `entry + cm.Entry[localFuncIdx]`.

```go
func runModuleI32(t *testing.T, m *wasm.Module, localFuncIdx int, args ...int32) int32 {
	t.Helper()

	cm, err := CompileModule(m)
	if err != nil {
		t.Fatalf("compile module: %v", err)
	}
	if localFuncIdx < 0 || localFuncIdx >= len(cm.Entry) {
		t.Fatalf("local function index %d out of range", localFuncIdx)
	}

	eng, err := runtime.NewEngine()
	if err != nil {
		t.Fatal(err)
	}
	defer eng.Close()

	jm, err := runtime.NewJobMemory(65536)
	if err != nil {
		t.Fatal(err)
	}
	defer jm.Close()

	ar, err := runtime.NewArena(4096)
	if err != nil {
		t.Fatal(err)
	}
	defer ar.Close()

	mem, entry, err := runtime.MapCode(cm.Code)
	if err != nil {
		t.Fatalf("map: %v", err)
	}
	defer runtime.Unmap(mem)

	serArgs := ar.Alloc(128)
	results := ar.Alloc(128)
	trap := ar.Alloc(8)

	for i, a := range args {
		binary.LittleEndian.PutUint32(serArgs[i*8:], uint32(a))
	}

	fn := entry + uintptr(cm.Entry[localFuncIdx])
	if err := eng.Call(fn, serArgs, jm.LinearMemory(), trap, results); err != nil {
		t.Fatalf("call: %v", err)
	}
	return int32(binary.LittleEndian.Uint32(results))
}
```

Then add a test that intentionally makes the callee populate all four pinned registers, while the caller has meaningful values in all four pinned locals:

```go
func TestPinnedLocalsSurviveInternalCallClobberingAllPinnedRegs(t *testing.T) {
	m := watToModule(t, `(module
		(func $clobber (param i32 i32 i32 i32) (result i32)
			local.get 0 local.get 1 i32.add
			local.get 2 i32.add
			local.get 3 i32.add)

		(func (export "f") (param $x i32) (result i32)
			(local $a i32) (local $b i32) (local $c i32)

			local.get $x i32.const 10 i32.add local.set $a
			local.get $x i32.const 20 i32.add local.set $b
			local.get $x i32.const 30 i32.add local.set $c

			i32.const 1 i32.const 2 i32.const 3 i32.const 4
			call $clobber
			drop

			local.get $x
			local.get $a i32.add
			local.get $b i32.add
			local.get $c i32.add))`)

	// x=5, a=15, b=25, c=35.
	// Without caller spill/reload around call, the callee would leave 1,2,3,4
	// in the pinned registers and this would return 10 instead of 80.
	if got := runModuleI32(t, m, 1, 5); got != 80 {
		t.Fatalf("pinned locals after internal call = %d, want 80", got)
	}
}
```

That test is important enough that I would make it a merge requirement.

## Add one stronger alias-capture test

The current alias test is good, but it only captures one pending self-read before `local.tee` overwrites the same local.

I would add a case with multiple pending reads plus a cross-pinned assignment. It exercises the `vPinned` path where the new value comes from another pinned local, not just from a constant:

```go
func TestPinnedAliasCaptureMultiplePendingReadsCrossPinnedSet(t *testing.T) {
	m := watToModule(t, `(module
		(func (export "f") (param $a i32) (param $b i32) (result i32)
			local.get $a
			local.get $a
			local.get $b
			local.tee $a
			i32.add
			i32.add))`)

	// old(a) + old(a) + b = 7 + 7 + 11.
	// If pending vPinned reads alias the overwritten register, this tends to
	// become 11 + 11 + 11 instead.
	if got := runI32(t, m, 7, 11); got != 25 {
		t.Fatalf("multi-alias capture = %d, want 25", got)
	}
}
```

## Smaller recommendations

I would tweak the `pinnedPool` comment. “Callee-clobbered scratch” is true for this engine’s generated-to-generated wasm ABI, but not for the normal Go/System V boundary. Since `RBX/R13/R14/R15` are only safe because the trampoline saves/restores them, the comment should say something like “clobbered by generated wasm callees under our wrapper ABI; preserved only by `enterNative` at the Go boundary.”

I would also add one `br_table` or conditional-branch test with pinned locals live across the target, because `opBrTable` has its own flush/move-slot path. The existing loop and if/else tests cover important join behavior, but `br_table` is a separate dispatcher path and cheap to cover.

## Merge recommendation

Request changes for test coverage, not for the core approach. After adding the internal-call clobber test and the stronger alias test, I would be comfortable merging this.

PR #27, opened 2026-06-29
feat(amd64): float rounding ops + copysign

JairusSW:
## What

Lowers the remaining float arithmetic ops (decoder/validator already handled them):

- **`f32/f64.ceil`, `floor`, `trunc`, `nearest`** (0x8d–0x90, 0x9b–0x9e)
- **`f32/f64.copysign`** (0x98, 0xa6)

This completes the f32/f64 arithmetic surface.

## Changes

- **Rounding** via SSE4.1 `ROUNDSS`/`ROUNDSD` with the matching rounding-mode immediate. `nearest` uses round-to-nearest-**ties-to-even** (matching wasm); bit 3 is set to suppress the precision (inexact) exception.
- **`copysign`**: `(a & ~signbit) | (b & signbit)` via `andps`/`andpd` + `orps`/`orpd`, reusing the existing sign-mask loader (factored out as `loadFMask`, now shared with `abs`/`neg`).
- **`asm_sse.go`** — new `Round` encoder. **`frontend.go`** — add the ten `InstrKind`s to the support allowlist.

> Note: `ROUNDSS`/`SD` require SSE4.1 (universal on x86-64 since ~2010); WARP's x86_64 backend uses the same instructions for these ops.

## Tests

`float_round_test.go` verifies every rounding mode including **ties-to-even** (`nearest 2.5→2`, `3.5→4`), signed inputs, `copysign` including a **negative-zero** sign source, and the f32 paths (driven via demote/promote). CLI: `ceil(2.1)=3`, `nearest(2.5)=2`, `copysign(4,-1)=-4`. Full `go test ./...` green.

## Performance

`BenchmarkCompile` (count=6): **neutral** — geomean −0.37% (noise), `Compile/float` unchanged, identical B/op and allocs. Purely additive; the corpus exercises none of the new opcodes.

---
Port #4 of the WARP singlepass porting effort. Independent of #24, #25, #26. The related float→int **trapping** truncation (still computes-but-doesn't-trap, per FEATURES.md) is left for a follow-up correctness port.


jtenner:
I reviewed PR #27. **I don’t see a clear semantic blocker in the lowering itself**: the opcode dispatch covers the expected f32/f64 rounding and `copysign` opcodes, the `ROUNDSS/ROUNDSD` encoder is straightforward, and `fcopysign` implements the expected `(a & ~signbit) | (b & signbit)` shape. The CI run for the head commit is green, and the bot report shows 614 passed / 2 skipped.

### Review comments I would leave

**1. Potential platform-correctness question: SSE4.1 is now required for modules using these ops.**
`Round` emits SSE4.1 `ROUNDSS/ROUNDSD`; there is no CPUID gate or fallback in this PR. That may be fine if the backend already assumes a modern x86-64 feature set—there are already non-baseline instructions like `LZCNT`, `TZCNT`, and `POPCNT` in the assembler—but it should be an explicit project decision. If the intended target is “any amd64 Linux,” this is a runtime illegal-instruction risk. If the intended target is modern amd64, I’d document that assumption in the backend/README/features notes.

**2. Strengthen tests to compare raw bits, not only float equality.**
The new rounding and `copysign` tests convert the raw result to `float64` and compare with `!=`. That will not distinguish `+0` from `-0`, which matters for these ops. I’d add bit-exact cases using `runF64Raw` directly. Good cases: `f64.trunc(-0.3) -> -0`, `f64.ceil(-0.3) -> -0`, `f64.nearest(-0.5) -> -0`, `f64.copysign(-0.0, +1.0) -> +0`, and the analogous f32 cases.

**3. Add NaN/payload tests for `copysign`.**
`copysign` is a bit operation, so it should preserve the magnitude/payload bits of the first operand and take only the sign bit from the second. The implementation does that, but the tests only cover ordinary finite values and a negative-zero sign source. I’d add raw-bit tests like `a = 0x7ff8000000001234`, `b = -0.0`, `want = 0xfff8000000001234`, plus an f32 version.

**4. Add an assembler-level test for `Round` high-register encoding.**
The new encoder has custom REX handling for `dst >= 8 || src >= 8`, but the WAT tests are simple functions that likely only allocate low XMM registers. A byte-level test for something like `Round(8, 9, false, roundCeil)` and `Round(10, 11, true, roundFloor)` would cover the REX branch directly.

**5. Add a frontend/support-pass regression test.**
The PR adds the ten instruction kinds to the frontend support allowlist, but the backend tests call `CompileFunction` directly, so they don’t necessarily prove the frontend/CLI path accepts these ops. A small frontend or CLI acceptance test with one rounding op and one `copysign` op would close that gap.

**6. Code reuse: name the float bit masks and consider folding constants.**
`fsign` already folds exact `abs`/`neg` bit operations for constants, and `foldFloatBin` has an established pattern of folding only when doing so won’t alter NaN behavior. `fcopysign` could be folded exactly even for NaNs because it is purely bitwise; `fround` could be folded for non-NaN constants using the same “don’t fold NaN results” policy. I’d also extract named constants for the f32/f64 sign and magnitude masks to avoid repeating magic literals across `fneg`, `fabs`, and `fcopysign`.

**7. Optional micro-cleanup: reduce `fcopysign` mask loading.**
Current `fcopysign` loads a magnitude mask, ANDs `a`, then reloads a sign mask and ANDs `b`. A one-mask form using the sign mask plus `andnps/andnpd` could avoid the second immediate load, though the current version is easier to read. I would only change this if the project prefers tighter codegen over clarity.

Overall: I’d be comfortable with the core lowering, with the caveat that the SSE4.1 requirement should be intentional/documented. The most valuable follow-up is bit-exact test coverage for signed zero and `copysign` NaN payload/sign behavior.

JairusSW:
Thanks for the thorough review — addressed in 7373ed4:

1. **SSE4.1 baseline** — documented as an explicit, intentional "modern amd64 (since ~2013)" assumption in `ARCHITECTURE.md`, alongside the existing LZCNT/TZCNT/POPCNT usage. No CPUID gate (matches the project's existing non-baseline instructions).
2. **Signed-zero bit-exact tests** — `TestFloatRoundingSignedZero` compares raw bits via `runF64Raw`: `trunc(-0.3)`/`ceil(-0.3)`→−0, `nearest(-0.5)`→−0, `copysign(+0,−1)`→−0, `copysign(−0,+1)`→+0, with f32 analogues.
3. **copysign NaN payload/sign** — `TestFloatCopysignNaN`: `a=0x7ff8000000001234, b=-0.0 → 0xfff8000000001234`, plus an f32 case via a new `runF32Raw` helper (direct, so the payload isn't reshaped by demote/promote).
4. **Assembler REX test** — `TestRoundEncoding` byte-checks `Round(R8,R9,…)` and `Round(R10,R11,…)` (the high-register REX branch), plus a no-REX low case.
5. **Frontend acceptance** — `TestDecodeValidateAcceptsFloatRoundingOps` exercises the support-pass/CLI path for the rounding + copysign kinds (backend tests use `CompileFunction` directly).
6. **Reuse/folding** — named `fSign*/fMag*` mask constants shared by neg/abs/copysign; `copysign` now folds const-const (purely bitwise → bit-exact incl. NaN/signed-zero, covered by `TestFloatCopysignFold`). I deliberately left **`fround` constant-folding out** — it would depend on Go's `math` matching SSE `roundsd` exactly, an avoidable divergence risk for a rare pattern.
7. **andnps micro-cleanup** — skipped, per your "only if the project prefers tighter codegen over clarity."

Full `go test ./...` green incl. `-race`.

PR #42, opened 2026-06-30
feat(amd64): register-based internal call ABI [WIP — step 1: parallel-move resolver]

JairusSW:
P1 from OPTIMIZATIONS.md — the register-based internal-call ABI that targets the **one workload wago loses to wazero** (recursive calls: FibRec 582µs wago vs 507µs wazero). **Stacked on #41.** Landing incrementally, default-OFF, so the main path/spec suite is never at risk.

## This step: RegisterCopyResolver (the prerequisite)
`regcopy.go` — a faithful Go port of WARP's `common/RegisterCopyResolver.hpp`. Placing call args into their ABI registers is a *parallel move*: targets may still hold values other moves need, and the dependency graph can contain cycles (RAX→RCX while RCX→RAX). `resolveRegMoves` emits every move whose target isn't a pending source, then breaks residual pure cycles with swaps.

Fully unit-tested (`regcopy_test.go`) via a simulated register file: independent moves, chains, 2-/3-cycles, cycle+chain, two disjoint cycles, and k-rotations for k=2..7.

## Remaining wiring (designed, next commits — all gated behind `CompileOptions.RegisterCallABI`, default off)
1. **Internal ABI:** linMem=RDI, trap=RSI; int args RAX,RCX,RDX,R8–R11; float args XMM0–7; result in RAX/XMM0. Functions whose signature fits (≤7 int / ≤8 float params, ≤1 result) get the register path; others fall back to the existing wrapper call.
2. **Dual entry per fitting function:** offset 0 = a thin host→internal *adapter* (marshals the serArgs buffer into arg regs, calls the internal entry, stores the register result back) so exports/host calls and `CompileFunction` keep working unchanged; the internal entry holds the real frame + body.
3. **Layout/reloc:** track each function's internal-entry offset; `callReloc` gains an `internal` flag so internal calls patch to the internal entry, wrapper calls to offset 0.
4. **Call site:** `emitRegisterCall` places args via `resolveRegMoves`, keeps the pinned-local spill/global write-back, does the trap check without clobbering the result register, pushes the register result.
5. Validate: FibRec correctness + full spec-exec suite **under the flag**, measure vs wazero, then flip the default.

The resolver here is correct and reviewable on its own; the rest builds on it.

### DIFF
--- a/bench/bench_test.go
+++ b/bench/bench_test.go
@@ -93,7 +93,7 @@ func BenchmarkInstantiate_wazero(b *testing.B) {
 func wagoSetup(b *testing.B, wasmBytes []byte, export string) (func(n int32) int32, func()) {
 	m, _ := wasm.DecodeModule(wasmBytes)
 	wasm.ValidateModule(m)
-	cm, err := amd64.CompileModule(m)
+	cm, err := amd64.CompileModuleWith(m, amd64.CompileOptions{RegisterCallABI: os.Getenv("WAGO_REG_ABI") == "1"})
 	if err != nil {
 		b.Fatal(err)
 	}
--- a/src/core/compiler/backend/amd64/asm.go
+++ b/src/core/compiler/backend/amd64/asm.go
@@ -112,6 +112,11 @@ func (a *Asm) MovReg64(dst, src Reg) {
 	a.emit(rex(true, src >= 8, false, dst >= 8), 0x89, 0xC0|((byte(src)&7)<<3)|byte(dst&7))
 }
 
+// Xchg64 exchanges the contents of two 64-bit registers (xchg r/m64, r64).
+func (a *Asm) Xchg64(x, y Reg) {
+	a.emit(rex(true, x >= 8, false, y >= 8), 0x87, 0xC0|((byte(x)&7)<<3)|byte(y&7))
+}
+
 func (a *Asm) Movsxd(dst, src Reg) {
 	a.emit(rex(true, dst >= 8, false, src >= 8), 0x63, 0xC0|((byte(dst)&7)<<3)|byte(src&7))
 }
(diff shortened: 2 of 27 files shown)


JairusSW:
## Complete — implemented, validated, measured

The full register ABI is now wired (still **opt-in**, default off).

**Result — flips the one benchmark wago was losing.** FibRec, wago vs wazero v1.9.0 (Ryzen 7 7800X3D):

| | wago | wazero | verdict |
|---|--:|--:|---|
| Wrapper ABI (old) | 586,051 ns | 511,380 ns | wago **15% slower** |
| **Register ABI (new)** | 489,413 ns | 534,932 ns | wago **~9% faster** |

−16.5% wago-internal on recursion, and −26% code size on fibrec (343→254 bytes), 0 allocs.

**What landed**
- `RegisterCopyResolver` (parallel-move solver) + unit tests.
- Internal ABI: linMem=RDI, trap=RSI, int args RAX/RCX/RDX/R8–R11, result RAX. Int-only signatures (≤7 params, ≤1 result) fit; float/multi-result fall back to the wrapper path transparently.
- Dual entry per fitting function: host→internal adapter at offset 0 (exports/`CompileFunction` unchanged) + internal entry with the real frame/body.
- Layout/reloc: `callReloc.internal` retargets internal calls to the internal entry.
- `emitRegisterCall`: args placed via the resolver, pinned-local/global save-restore kept, trap check avoids the result register.
- Threaded through `RuntimeConfig` (`WithRegisterCallABI`, or `WAGO_REG_ABI=1`); `compileFunc` prologue/body refactored into shared helpers so the wrapper path is byte-identical.

**Validation:** full spec-exec suite + all package tests pass under `WAGO_REG_ABI=1`; integration tests cover recursion, leaf-via-adapter, multi-arg placement, and float fallback.

**Finding worth noting:** arg-marshalling was *not* the only cost — at PinNone the register path is −19% (vs −16.5% at PinHotness), i.e. pinning a local that's live across calls still costs a spill/reload pair per call site. The deferred "−10 live-across-calls" pinning penalty (P2 follow-up) is the next lever.

**Default flip:** kept off in this PR (conservative for an ABI change); recommend flipping `RegisterCallABI` on by default in a follow-up after soak, since the spec suite already passes under it.

PR #69, opened 2026-07-02
perf(railshot): return single int result in RAX (WARP target hint, step 4)

JairusSW:
Carries out the first slice of **step 4 (target hints)** from Codex's WARP x64 port plan (`docs/warp-x64-port-plan.md`): produce a function's single integer result **directly in RAX** at every exit, instead of the flush-to-slot-0 + epilogue-reload round trip.

## What

For functions using the register-return ABI with exactly one integer result (`f.singleIntResult`), each exit — fall-through end, `return`, and br/br_if/br_table to the function frame — places the result in RAX (the WARP-style return target hint), and the reg-ABI epilogue no longer reloads it from slot 0.

Gated on **`!touchesMemory`**: the hint helps compute/call-heavy code but adds register pressure in deep memory-bound call graphs (json-as's TLSF/GC), where it measured as a small regression — so it only fires where it's a win. Multi-result, float-result, and wrapper-ABI functions are unchanged.

## Results (guard mode, interleaved A/B, n=8, wazero control confirms no drift)

| workload | before | after | Δ |
|---|--:|--:|--:|
| ExecFibRec (recursion) | 423.4µs | 368.7µs | **−12.9%** (p=0.000) |
| ExecFibLoop | 19.54ns | 19.31ns | −1.1% |
| Exec/fib_iter | 38.70ns | 37.47ns | −3.2% |
| Exec/dispatch | 25.09ns | 24.77ns | −1.3% |
| Exec/memory_tree | — | — | neutral |
| JsonAsSerialize | 279.4ns | 280.4ns | ~ (p=0.37) |
| JsonAsDeserialize | 454.8ns | 453.7ns | ~ (p=0.20) |
| JsonAs* wazero (control) | — | — | ~ (p=0.86/0.62) |

Recursion −12.9%; no regression on the memory-bound json-as (the plan's don't-regress target) or memory_tree. Meets the plan's acceptance bar (no >3% regression on json-as-guard / memory_tree-guard / fib_rec).

## Validation

- Spec suite **1.0 (15,346 assertions), 2.0, 3.0** all pass.
- `go test ./...` green; railshot package tests green; gofmt clean.
- AS module correctness re-checked (blake `hashN(100)=-1321215924`, json `ser(200)=21600`, utf `conv(200)=819200`).

Remaining step-4 sites (select, call-params) and the larger §B item (operand-stack-in-registers across branches) are deliberately left out — they need model restructuring, not simple hint additions, and per the status doc are low-ROI for the memory-bound json-as target.

### DIFF
--- a/src/core/compiler/backend/railshot/compile.go
+++ b/src/core/compiler/backend/railshot/compile.go
@@ -50,6 +50,14 @@ type fn struct {
 	addRspAt  int  // byte offset of the epilogue's AddRsp imm32 (patched with frameSize)
 	guardMode bool // elide inline bounds checks; rely on guard-page + SIGSEGV trap
 	lazyZero  bool // defer declared-local zeroing for small call+memory functions
+	// singleRegResult: this function uses the register-return ABI with exactly one
+	// result. Its exits produce that result directly in the return register — RAX
+	// (int) or XMM0 (float) — via the WARP-style target hint, skipping the
+	// flush-to-slot-0 + epilogue-reload round trip. resultFloat/resultF64 cache the
+	// result's type for that placement.
+	singleRegResult bool
+	resultFloat     bool
+	resultF64       bool
 
 	// Control-flow state (Phase 3).
 	ctrl        []ctrlFrame // open block/loop/if frames; ctrl[0] is the function frame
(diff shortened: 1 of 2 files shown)


JairusSW:
Extended to cover **single float results (XMM0)** too, not just int (RAX) — completes the `returns` hint site for both register classes. Same gate (`!touchesMemory`, reg-ABI, single result). Validated: spec suite 1.0/2.0/3.0 all pass, `go test ./...` green. Int-return benchmarks unchanged by the float path (fib_rec still −12.9%); float returns are rarer so no dedicated bench, but the spec suite exercises float function returns extensively.

PR #70, opened 2026-07-02
perf(railshot): operand-stack-in-registers, phase 2 — register-merge single-int blocks (flag-gated)

JairusSW:
First implementation step of the operand-stack-in-registers refactor (`docs/operand-stack-registers-plan.md`) — WARP's RegisterCopyResolver model, replacing the flush-to-slot reconciliation at control boundaries.

## What (phase 2, flag-gated)

A plain `block` producing exactly **one integer result** carries that value in a canonical register (`mergeReg` = RBP) across **all** its edges instead of a frame slot:
- **fall-through** end: result computed straight into `mergeReg` (no slot store).
- **br / br_if / br_table / fused-br_if** edges: result loaded into `mergeReg` from its canonical slot (slot copy kept intact so a `br_if` fall-through still sees the value).
- **merge landing**: below-operands rebuilt as slots `[0,height)`, result pushed as a register-resident value.

Excludes loops (params/back-edge), `if` (else/passthrough edges), floats, and multi-value — those are later phases. Gated behind `WAGO_REG_MERGE` (**default off**), so main behavior is unchanged; the slot path stays as the reference oracle.

## Validation

- **Correctness (flag on):** spec suite **1.0 (15,346), 2.0, 3.0** all pass (value-checked, not just no-trap); `go test ./...` green.
- **Differential:** every corpus + AS-module exec entry (18) produces **identical** results flag-off vs flag-on.
- **Path is real:** committed `TestRegMergeBlockResult` + a crafted module (`block (result i32)` via br_if / br_table / nesting / result-as-operand) confirm the reg path *fires* (codeLen 528→497, fewer slot round-trips) and matches the slot path exactly.

## Scope note

Single-result `block`s don't appear in the current corpus (AS/wat control flow uses `if`/`loop`), so this is **inert on the corpus today** — it lands and validates the reconciliation machinery (`reconcileMerge1`, `branchEdgeToMerge1`, merge landing, all edge types) that phases 3+ (`if`/`else`, multi-value, below-operands-in-registers, loops) build on. Default stays off until a phase covers the constructs real code uses; then flip after a soak.

### DIFF
--- a/docs/operand-stack-registers-plan.md
+++ b/docs/operand-stack-registers-plan.md
@@ -1,0 +1,89 @@
+# Operand-stack-in-registers refactor (WARP RegisterCopyResolver)
+
+## Goal
+
+Replace railshot's **canonical-slot** reconciliation at control boundaries with
+**register** reconciliation — WARP's model. Today, at every block / if / loop /
+branch / end, `flush()` writes the whole operand stack to position-indexed frame
+slots (`spillOff(i)`) and `setDepth` rebuilds it as slot entries; merges are
+trivial because all edges agree on slots. This costs a store+reload per operand
+per boundary. WARP instead keeps operands in registers across merges and shuffles
+them into a canonical register assignment at each edge via a parallel-move
+resolver. Target win: call/branch-heavy code (`memory_tree`), not json-as
+(memory-bound; leave it alone).
+
+Infra already present: `regcopy.go` `resolveRegMoves` (parallel move with cycle
+handling), used today for call args. Reuse it for edge reconciliation.
+
+## Current model (facts)
+
+- `flush()` (control.go): each operand i → `spillOff(i)`, deferred nodes
+  condensed, all registers freed, stack rebuilt as `stSlot` entries. Idempotent
+  for already-canonical slots.
+- `moveSlots(from,to,n)`: copy n slots (used to place branch values at the target
+  frame's `height`).
+- `ctrlFrame{height, paramN, resultN, branchN, ...}`: `branchN` = values
+  transferred to this label (results for block/if, params for loop).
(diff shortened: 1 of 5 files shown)


JairusSW:
## Phase 3 added — if/else register-merge, default ON

Extended the reg-merge to `if`/`else` single-int results (the common value-producing construct): then-edge (opElse), else fall-through, br/br_if/br_table edges, and the no-else cond-false passthrough stub. **Flipped default ON** — `WAGO_REG_MERGE=0` keeps the slot path as the reference oracle.

### Perf (guard mode, interleaved n=8, wazero control = neutral → clean)
| workload | off | on | Δ |
|---|--:|--:|--:|
| **Exec/fib_rec** (recursion + if/else) | 1.629 ms | 1.406 ms | **−13.7%** (p=0.000) |
| JsonAsSerialize (wago) | 270.3 ns | 266.2 ns | −1.5% (p=0.001) |
| memory_tree | 11.97 µs | 11.93 µs | −0.27% |
| fib_iter / dispatch / branches / mandelbrot / sieve / json-deser | — | — | neutral |

fib_rec (`if (result) then n else fib(n-1)+fib(n-2)`, ~2²⁸ calls) is the headline. No regressions.

### Correctness
- spec **1.0 (15,346) / 2.0 / 3.0** pass with default (now on) and with `=0`.
- Full corpus + AS-module differential: **18/18 exec results identical** off vs on.
- Committed `TestRegMergeBlockResult` + `TestRegMergeIfElse`; crafted modules (abs / nested sign / if-in-loop / br_if / br_table) verified.

Remaining phases (multi-value, below-operands-in-registers, loops) tracked in the plan doc.

PR #81, opened 2026-07-02
test(spec): exclude out-of-scope files from the 1.0 scoreboard (54/54, 0 fails)

JairusSW:
The MVP scoreboard counted files testing features wago deliberately doesn't implement — multi-instance `linking`, the spectest host module's imported `memory`/`data`, non-numeric host `imports`, and imported `start` — as failures/blocked against the denominator, dragging it to 54/58 (93%) even though every one of those is a missing *feature*, not a miscompile.

This adds an `outOfScope` set: those files are still run and shown in the table (as **N/A** with a reason) but excluded from the conformance denominator and assertion tallies.

Result: **54/54 applicable files fully passing · assertions pass=16476 fail=0** — i.e. zero miscompiles across the in-scope 1.0 core. Matches `make spec1` (TestSpecSuiteExec), which already skips these. Drop an `outOfScope` entry if wago gains the capability.


JairusSW:
Reverting: per feedback, the 1.0 MVP scoreboard should stay honest at <100% rather than excluding linking/imports as N/A. N/A is reserved for pre-phase-1 proposals only.

PR #88, opened 2026-07-02
perf(railshot): backlog sweep — trap stubs, strength reduction, scaled-index LEA, lazy merges, jump tables, reg-ABI call_indirect, const bulk-mem

JairusSW:
The full perf/overhead backlog from the #87 investigation plus the still-open items from OPTIMIZATIONS.md (which this PR also rewrites to match reality). Eight changes, each gated on the spec suite + full tests + the explicit-vs-guard differential corpus, and benchmarked individually:

1. **Shared cold trap stubs (P9)** — every bounds/div/indirect/fence check was jumping over a ~20-byte inline trap block; now one stub per trap code per function and hot checks fall through (`ja stub`). json-as module code −23%; sieve −25%, memory_tree −19% on their own.
2. **Leaf fence skip + one-pass pre-scan + guard default** — call-free leaves with provably small frames skip the stack-fence check (the 256 KiB margin absorbs one unchecked frame); `hints.go`'s seven AST walks fused into one `scanBody`; `-tags wago_guardpage` builds now default to guard bounds (`WAGO_BOUNDS=explicit` overrides).
3. **Algebraic identities + strength reduction (P4)** — `x±0`, `x&~0`, `x*2ⁿ→shl`, `x*{3,5,9}→lea`, `x/ᵤ2ⁿ→shr`, `x%ᵤ2ⁿ→and`, `x^x→0`, … applied at `pushBinOp` before a deferred node exists. 24 exec tests incl. unsigned edge values.
4. **Scaled-index LEA fusion** — `add(x, shl(y,k≤3))` → `lea [x+y*2ᵏ]` (the AS array-address shape). linked_list −17%.
5. **Lazy per-frame merge agreement (P6, locals)** — control edges record/converge per-frame pinned-local states, so call-clobbered locals can stay slot-only across branch-dense merges until read; loop tops stay eager (reloads hoisted); conditional returns converge nothing. −17% reload traffic on json-as. Exec tests cover both-edges-lazy, asymmetric-both-ways, no-else stubs, cond-returns.
6. **Register-ABI `call_indirect` + `call;local.set` fusion** — table entries carry the internal-entry delta in their pad word (codec v9), so type-checked compatible signatures bypass the wrapper adapter; a register call's result lands directly in a pinned local's register. dispatch −8%.
7. **`br_table` jump tables (P7)** — n≥5 dispatches through a RIP-relative offset table with deduplicated case stubs (new `JmpReg`/`LeaRipPlaceholder` encoder ops).
8. **Small constant `memory.fill`/`copy` unrolled** — n≤32 lowers to overlap-safe load-all/store-all chunks, preserving memmove semantics (25 exec tests incl. both overlap directions). blake −4%.

Also: internal reg-ABI entries and loop tops are 16-byte aligned (tight-loop benchmarks swing ±20% on layout luck otherwise — sieve and fib_rec both demonstrated this during the work).

## Results (explicit bounds, vs main/#87)

| bench | main | this PR | Δ |
|---|---:|---:|---:|
| sieve | 163µs | 123µs | **−24%** (now beats wazero) |
| memory_tree | 14.6µs | 11.8µs | **−19%** |
| linked_list | 11.3µs | 9.4µs | **−17%** |
| dispatch | 19.1ns | 17.6ns | −8% |
| blake-as | 729µs | 700µs | −4% |
| json-as ser/deser | 218/396 | 214/380 | −2%/−4% |
| memory.sum | 337ns | 230ns | **explicit == guard mode** |

Cumulative including #87: json ser 257→214, deser 420→380; memory.sum 552→230 (−58%); sieve −25%; memory_tree −32%.

## Validation
- Spec suite: unchanged (all files pass except pre-existing linking/data import gaps); br_table 146/146, call_indirect 122/122
- `go test ./...` green; ~60 new exec-test assertions across the new paths
- Explicit-vs-guard differential corpus (blake/utf/json/memory_tree/sieve/memory): all match
- `.wago` codec bumped to v9 (InternalEntry); malicious-count reader tests extended

OPTIMIZATIONS.md rewritten: P1/P2/P3/P4/P5/P6/P7/P9 and the MVP batch marked landed with measurements; remaining roadmap is vFlags (P8), float parity, the deser/SSA question, WARP runtime infra (cancel, stack traces, arm64, V2 imports), and CodegenStats.

### DIFF
--- a/OPTIMIZATIONS.md
+++ b/OPTIMIZATIONS.md
@@ -4,296 +4,140 @@ Two complementary lenses on the same question — *how do we make wago faster wi
 destroying the reason it exists* (fast compile, no cgo, tiny footprint, single pass):
 
 1. **Make the single-pass backend smarter** — better-informed choices inside the existing
-   Valent-Block tier (the bulk of the actionable near-term work).
-2. **Port what's still worth porting from WARP** (`warp/`) — the C++ reference engine wago
-   is a port of. Used here as a *reference axis*, not a target to clone.
+   railshot tier.
+2. **Port what's still worth porting from WARP** (`warp/`) — the C++ reference engine the
+   backend is a port of. Used as a *reference axis*, not a target to clone.
 
 The headline architectural decision (see the end): **keep two tiers and do not blend them.**
 The single-pass tier stays single-pass; the `src/core/compiler/ir` SSA package becomes an
 *optional* optimizing tier later, never something a plain `Compile` pays for.
 
-Legend: effort S/M/L · value ⬜ low · 🟦 medium · 🟩 high · ⭐ very high ·
-status ✅ done · 🚧 partial · ⬜ todo. All wago citations verified against current source.
+Legend: effort S/M/L · value ⬜ low · 🟦 medium · 🟩 high · ⭐ very high.
 
 ---
 
-## What's already in place
+## What's in place (updated 2026-07-02)
 
-The backend (`src/core/compiler/backend/railshot`) is a single-pass x86-64 codegen fused with
-register allocation over a symbolic operand stack (`vConst`/`vLocal`/`vReg`/`vSpill`/`vPinned`),
-flushing to deterministic frame slots at control-flow joins. Already done:
+The backend (`src/core/compiler/backend/railshot`) is the full WARP-architecture port: a
+single-pass x86-64 codegen over a valent-block operand stack (deferred-action trees,
+condense engine) with an on-the-fly whole-register-file allocator. Landed, in rough order:
 
-- **Immediate compare/branch fusion** — a compare/`eqz` keeps EFLAGS live for exactly one step
-  and fuses into an adjacent `if`/`br_if`/`select`, skipping `setcc`+`test` (`fuse.go:26`).
-- **Constant folding** — integer arith, shifts, compares, unary, safe div/rem, some float
-  (`fold.go`).
-- **Lazy + pinned locals** — `local.get` stays symbolic (`vLocal`); first few integer locals
-  pinned to `RBX/R13/R14/R15` (`compile.go:38-43,81-82`).
-- **Pinned mutable integer globals** — lazy write-back, using pinned slots left after locals
-  (`compile.go:45-53,1164-1228`).
-- **Memarg offset folding** — constant memarg offset folded into the addressing mode (PR #36,
-  `−17%` on an offset-heavy sum, bounds check preserved).
-- **Experimental guard-page bounds-check elision** — `4096`-load array sum `3566 → 2686 ns/op`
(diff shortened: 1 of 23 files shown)


JairusSW:
**Follow-up pushed to this branch: WARP-replication batch.** Ran WARP itself on json-as (its Linux default = passive bounds, so compared against wago guard mode): WARP ser 97ns / deser 164ns per unit vs wago's 190/338 at the time — 2× on both. Diffed hot functions instruction-by-instruction (wago's TLSF `removeBlock` is now *smaller* than WARP's — 92 vs ~100 instructions) and replicated the structural edges that remained:

1. **No per-call environment protocol** (`73d0a51`) — RBX/linMem is now a whole-module invariant (adapter establishes it at the Go boundary; wasm→wasm calls pass nothing), and the trap-cell pointer lives in basedata (`abi.TrapCellPtrOffset`, installed once per entry by `Engine.Call`, which also zeroes the cell). Deletes 6 instructions + trap-slot traffic per internal call round-trip, including the per-return trap-clear. fib_rec −13%.
2. **Borrowed load addresses + read-only branch conditions** (`e38616d`) — deferred integer loads index `[rbx + pinned_reg]` directly (realize-on-`local.set` keeps it sound); `test` conditions read pinned locals in place. sieve −23% (95µs).
3. **Module-wide global register pinning** (`e600dba`) — the hottest mutable int global (AS shadow-stack pointer) lives in R14 in *every* function: registers cross wasm→wasm calls untouched, cells sync only at offset-0 boundaries, wrapper calls, and trap stubs. K=1 after measuring K=2/3 (pin-pool pressure hurts blake).
4. **Inline chunk-loop memmove for small dynamic copies** (`42c05c5`) — the deserialize breakthrough: AS `__renew` (47% of deser!) makes thousands of small dynamic `memory.copy`s, each paying `rep movsb` startup. n<96 now runs WARP's 8-byte chunk loop (overlap-safe both directions); large sizes keep rep/ERMSB. **Deserialize −43%.**

## json-as, ns per unit (guard/passive)

| | ser | deser |
|---|---:|---:|
| WARP | 97 | 164 |
| wazero | 141 | 293 |
| wago before this batch | 204 | 342 |
| **wago now** | **189** | **185** |

**wago deserialize now beats wazero by 1.43× and is within 1.13× of WARP.** Serialize remains ~2× (52% in one bump-pointer store-burst function — analysis in OPTIMIZATIONS.md R4).

Validation unchanged and green throughout: spec suite, `go test ./...`, explicit-vs-guard differential corpus, 40 new dynamic bulk-mem exec tests (overlap both directions, boundary sizes).

JairusSW:
The WARP-replication follow-up described above has been split out into #89 (stacked on this branch) — this PR is back to ending at the OPTIMIZATIONS.md refresh (`8604c1e`).

PR #96, opened 2026-07-02
Compile through no-body wasm validation

jtenner:
## Summary

`DecodeModule` no longer materializes a structured instruction tree for function
bodies. Decoded functions keep their compact `Locals` plus the raw expression
bytes (`BodyBytes`), and validation type-checks by walking that bytecode
directly instead of walking an in-memory AST. Const-expressions, element
segments, data offsets, and global initializers are likewise kept as
byte-backed summaries rather than decoded instruction trees.

Railshot compilation consumes the same validated bytecode, so a production
compile does **not** build or verify IR unless a codegen path explicitly asks
for it. Bytecode immediate skipping/classification is centralized in the `wasm`
package (`SkipInstructionImmediate` / `ClassifyInstructionImmediate`) and reused
from railshot and the frontend so the parsers cannot drift.

Validation state is reused across a module rather than reallocated per function:
resolved function/composite types are memoized by flat type index, the operand/
control stacks and byte cursor are reused across bodies, and single-value block
signatures and the scratch instruction payload avoid per-instruction
allocation. Together these keep the validator's allocation count near-flat in
module size.

## AST validator vs. byte-backed validator

| Aspect | AST validator (before) | Byte-backed validator (this PR) |
| --- | --- | --- |
| Function bodies | Decoded into a full `Instruction`/`Expr` tree at `DecodeModule` time | Kept as raw `BodyBytes` + compact `Locals`; no tree built |
| Const-exprs, elements, data, globals | Materialized as instruction trees | Byte-backed summaries, walked on demand |
| Type-checking | Walks the retained in-memory AST | Streams the validated bytecode directly |
| Live instruction state | Whole module's AST retained decode → validate → compile | ~one transient instruction at a time |
| Compile input | The decoded AST | The same validated bytecode, consumed directly by railshot |
| IR build/verify | Implicit in the decoded tree | Not built or verified unless a codegen path consumes IR |

## Performance

`DecodeModule` + `ValidateModule` over real `bench/corpus` modules (Go 1.22,
single core, mean of 3 runs, `-benchmem`):

| Module (decode + validate) | AST validator | Byte-backed (this PR) | Change |
| --- | --- | --- | --- |
| **wasm3** (184 KB) | 9.41 ms · 15.1 MB · 44,261 allocs | 6.04 ms · 300 KB · 1,604 allocs | −36% time · −98% mem · −96% allocs |
| **json-as** (23 KB) | 1.18 ms · 2.11 MB · 7,044 allocs | 0.68 ms · 47.7 KB · 280 allocs | −42% · −98% · −96% |
| **isa_mem** | 123 µs · 291 KB · 336 allocs | 110 µs · 6.47 KB · 57 allocs | −11% · −98% · −83% |
| **linked_list** | 7.13 µs · 14.9 KB · 70 allocs | 5.77 µs · 3.47 KB · 31 allocs | −19% · −77% · −56% |

The largest win is allocation: the AST validator retained a full instruction
tree for every function (~15 MB and ~44k allocations to decode+validate wasm3),
while the byte-backed validator holds only the raw bytes and one transient
instruction at a time and reuses its validation state across functions —
cutting allocations by roughly 25–30× on the larger modules.

## Correctness

The byte-backed path is validated as a differential oracle against the AST path:

- `FuzzDecodeValidateByteBackedDifferentialGenerated` — byte-backed
  decode+validate must agree with AST decode + `ValidateModule` on
  accept/reject and error phase (fuzzed for millions of executions, no
  divergence).
- Direct `SkipInstructionImmediate` / `ClassifyInstructionImmediate` coverage
  for the GC/bulk/SIMD/atomic immediate edge cases (`0xfb`–`0xfe`), including
  `br_on_cast`/`br_on_cast_fail` cast flags and `ref.test` heaptype handling.
- Official WebAssembly testsuite (`WAGO_SPECTEST_DIR`, `TestSpecSuite` /
  `TestSpecSuiteExec`): valid modules accepted, `assert_invalid` /
  `assert_malformed` rejected, all execution assertions pass.
- `go test ./...`.

## Non-goals kept out of scope

- No `validate-direct` CLI alias was added or restored.
- No production function-body AST decoding was reintroduced.
- No production IR build/verify gate was added.
- No frontend/runtime feature support was broadened beyond existing policy.


JairusSW:
i'll review tomorrow

PR #109, opened 2026-07-03
chore: ignore dirty content in warp submodule

JairusSW:
`warp/` is a reference submodule (upstream `wasm-ecosystem/wasm-compiler`) that accumulates local build output (`build-bench/`) and local bench instrumentation (`bench/main.cpp`), so `git status` was permanently noisy.

Set `ignore = dirty` so working-tree changes in the submodule are silenced, while still reporting if the pinned commit ever changes.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -2,6 +2,7 @@
 	path = warp
 	url = https://github.com/wasm-ecosystem/wasm-compiler
 	branch = develop
+	ignore = dirty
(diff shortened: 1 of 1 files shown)


JairusSW:
Superseded by the warp fork repoint (#110): the instrumentation is now a real commit on wago-org/warp, so the ignore=dirty workaround is unnecessary.

PR #129, opened 2026-07-04
feat(cli): --no-bounds-facts flag; thread P6.1 elision through RuntimeConfig

JairusSW:
Makes the P6.1 straight-line bounds-check elision (#128) a first-class, CLI-controllable option instead of only the `WAGO_NO_BOUNDS_FACTS` env var.

## Plumbing (config → codegen)

- **`RuntimeConfig.WithBoundsFacts(enabled bool)`** + `BoundsFacts()` — on by default (the field is `noBoundsFacts`, so the zero value is elision-on). Immutable-style like the other `With*`.
- **`amd64.CompileOptions.NoBoundsFacts`** — threaded to a new `fn.boundsFacts` field; `memAddr` checks `f.boundsFacts` instead of the package global. Effective elision = `boundsFactsEnabled (env) && !opts.NoBoundsFacts` — env var still works as a global override, and the per-instance link-time recompile caches the decision (`Compiled.noBoundsFacts`) like `boundsElide`.
- **`api.go`** passes `cfg.noBoundsFacts` through on both the initial compile and the link recompile.

## CLI

`wago run` gains `--no-bounds-facts` (bare boolean flag; `extractOpts` now supports boolean flags). `mustLoad` compiles raw wasm via `CompileWithConfig(cfg, …)` (a precompiled `.wago` already has its codegen baked in, so the flag only affects fresh compiles). Documented in `wago help`.

## Verification

- End-to-end through the built CLI: `wago run -e step nbody.wasm 2000` = `-169071606` **identically** with and without `--no-bounds-facts` (correctness), and the flag genuinely toggles codegen — nbody compiles to **4873 B with elision vs 5849 B without** (~17% smaller).
- New tests: `TestWithBoundsFacts` (config toggle + immutability + the option shrinks emitted code through `CompileWithConfig`), and a `NoBoundsFacts` case in `TestBoundsFactsElision` (option off → both checks emitted).
- Full battery: `go test ./...` (root + bench, both tag modes), `make test-guard`, spec 57/57 (pass=16592 fail=0 skip=1591). Correctness of the elision itself is covered by #128's differential-both-modes gate.

### DIFF
--- a/cli/wago/main.go
+++ b/cli/wago/main.go
@@ -60,6 +60,7 @@ func usage(w *os.File) {
 
 %s
   -e, --invoke <name>       export to call
+      --no-bounds-facts     disable straight-line bounds-check elision (explicit mode)
 
 %s
   wago add.wasm 2 3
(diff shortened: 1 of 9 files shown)


JairusSW:
Renamed the user-facing surface per feedback — `elide`/`bounds-facts` was compiler jargon:
- config: `WithDeferBoundsChecks(bool)` / `DeferBoundsChecks()` (was `WithBoundsFacts`)
- CLI: `--no-defer-bounds-checks` (was `--no-bounds-facts`), help text: "bounds-check every access (don't skip redundant checks)"

Railshot's internal `CompileOptions.NoBoundsFacts` / `WAGO_NO_BOUNDS_FACTS` keep the plan's term (not user-facing). Behaviour and gates unchanged; `wago run --no-defer-bounds-checks` verified identical result + toggles codegen.

PR #136, opened 2026-07-04
amd64: implement SIMD and relaxed SIMD railshot support

jtenner:
## Summary
- add amd64 x86 SIMD encoder primitives and golden tests needed by railshot lowering
- make `v128` a first-class railshot machine type for params, locals, stack values, spills, results, memory traffic, calls, and explicit unsupported cases
- implement the currently admitted core SIMD and deterministic relaxed SIMD lowering, including relaxed dot products without AVX2/FMA/VNNI requirements
- document the linux/amd64 SIMD baseline: SSE4.1 plus AVX/VEX.128 XMM encodings; no silent AVX2/FMA/VNNI/wider-vector dependency

## Validation
- `go test ./...`


src/core/compiler/backend/railshot/simd.go
```diff
@@ -0,0 +1,2185 @@
...
+	xa := f.materializeV128(a)
+	f.fpinned = f.fpinned.add(xa)
+	xb := f.materializeV128(b)
+	f.fpinned = f.fpinned.add(xb)
+
+	// PACKUSWB saturates signed 16-bit sources, but wasm's _u narrow first
+	// interprets each source lane as unsigned. Clamp unsigned words to 0x00ff
+	// before packing, so high-bit lanes (0x8000..0xffff) become 0xff instead of
+	// x86's signed-negative-to-zero result.
+	max := f.v128ConstReg(0x00ff00ff00ff00ff, 0x00ff00ff00ff00ff)
```
JairusSW: **Correctness bug.** The premise here is inverted. Wasm `i8x16.narrow_i16x8_u` interprets each source lane as a **signed** i16 and saturates it into the unsigned u8 range `[0,255]` — negative inputs saturate to `0`. That is *exactly* what `PACKUSWB` already does. The added `VPMINUW` pre-clamp treats the source as unsigned, so a negative lane like `0xFFFF` (-1) becomes unsigned `65535`, clamps to `0x00FF`, and `PACKUSWB` then emits `255` instead of the correct `0`.

Same issue in `i16x8NarrowI32x4U` below (`VPMINUD` + `PACKUSDW`).

Fix: drop both helpers and lower directly, matching the signed variants:
```go
case 102: // i8x16.narrow_i16x8_u
    f.v128Bin(f.a.VPpackuswb)
case 134: // i16x8.narrow_i32x4_u
    f.v128Bin(f.a.VPpackusdw)
```
This fixes all 25 failing `simd_conversions` assertions and also removes 3 instructions + a const materialization per op.

src/core/compiler/backend/railshot/simd_exec_test.go
```diff
@@ -0,0 +1,2324 @@
...
+
+	cases := []struct {
+		name string
+		a    [16]byte
+		b    [16]byte
+		sub  uint32
+		want [16]byte
+	}{
+		{"i8x16.narrow_i16x8_s", i16x8Bytes(200, -200, 127, -128, 0, 50, -50, 30000), i16x8Bytes(-32768, 32767, 1, -1, 128, -129, 100, -100), 101, i8x16Bytes(127, -128, 127, -128, 0, 50, -50, 127, -128, 127, 1, -1, 127, -128, 100, -100)},
+		{"i8x16.narrow_i16x8_u", i16x8Bytes(-1, -128, 0, 1, 254, 255, 256, 32767), i16x8Bytes(-32768, -2, 2, 300, 1000, 255, 256, -1), 102, i8x16Bytes(-1, -1, 0, 1, -2, -1, -1, -1, -1, -1, 2, -1, -1, -1, -1, -1)},
```
JairusSW: This expected value is wrong and masks the bug in the lowering. Input lane 0 is `-1` (a signed i16); `i8x16.narrow_i16x8_u` must saturate it to `0`, not `0xff`. Correct expectation:
```go
i8x16Bytes(0, 0, 0, 1, -2, -1, -1, -1, 0, 0, 2, -1, -1, -1, -1, 0)
```
Likewise line 1732 for `i16x8.narrow_i32x4_u` should be `i16x8Bytes(0, 0, 0, -1, -1, -1, 42, 0)`. Worth adding negative-input cases here explicitly — they are the ones PACKUS* handles differently from a naive unsigned clamp.

src/core/compiler/backend/railshot/simd.go
```diff
@@ -0,0 +1,2185 @@
...
+	xa := f.materializeV128(a)
+	f.fpinned = f.fpinned.add(xa)
+	xb := f.materializeV128(b)
+	f.fpinned = f.fpinned.remove(xa)
+	op(xa, xa, xb)
+	f.releaseF(xb)
+	f.pushVReg(xa)
+}
+
+func (f *fn) v128FloatMinMax(f64, isMax bool) {
```
JairusSW, line 223: **Optimization (not blocking).** This per-lane scalar extract/min/insert loop is ~30 instructions and the heaviest common vector op in the profile (~9.4 ns/op for f32x4). The standard branchless vectorized sequence reproduces Wasm NaN + signed-zero semantics in ~7 instructions (~7.7 ns/op, -18%):
```
t  = MINPS(a,b);  u = MINPS(b,a)         // (MAXPS for max)
c  = CMPUNORDPS(t,u)                      // NaN lane mask
t  = t OR u                              // ORPS for min (-0 wins); ANDPS for max (+0 wins)
t  = t OR c                              // force NaN lanes on
c  = PSRLD(c, 10)                        // PSRLQ 13 for f64
t  = ANDNPS(c, t)                        // canonicalize NaN lanes, keep the rest
```
I validated this against all 772 (f32x4) + 793 (f64x2) spec min/max assertions, NaN/±0 edges included. Happy to hand over the patch (needs two new immediate-shift encoder helpers, `VPsrldImm`/`VPsrlqImm`).

JairusSW (changes requested):
Lgtm

src/core/compiler/frontend/frontend.go
```diff
@@ -833,27 +910,49 @@ func (p supportPass) blockType(bt wasm.BlockType, context string) error {
...
+
+func supportedFrontendValType(v wasm.ValType) bool {
 	if v.Kind == wasm.ValNum {
 		switch v.Num {
 		case wasm.NumI32, wasm.NumI64, wasm.NumF32, wasm.NumF64:
-			return nil
+			return true
 		}
 	}
+	return v.Kind == wasm.ValVec && wasm.EqualValType(v, wasm.V128)
```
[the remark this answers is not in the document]
jtenner: Fixed in https://github.com/wago-org/wago/commit/5c194bd234bafc1382d4db366bcf6be02b1112b7.

What changed:
- Public signatures now expose `ValV128` instead of falling through to `ValI32`.
- `Invoke`/instantiate call-buffer sizing now count `v128` as two adjacent `uint64` slots (16 bytes) for both params and results.
- `Invoke` returns each `v128` result as two little-endian `uint64` slots and requires two input slots for each `v128` param.
- Added `TestInvokeV128UsesTwoPublicSlots` for scalar slots around a `v128` param/result.

Validation: `go test ./...` passes after pushing.

src/core/compiler/backend/railshot/simd.go
jtenner: Fixed in https://github.com/wago-org/wago/commit/3d9f2cfc702f371deaa42191617b33a77d5e35cb.

What changed:
- Removed the unsigned `VPMINUW`/`VPMINUD` pre-clamp helpers.
- `i8x16.narrow_i16x8_u` now lowers directly to `VPpackuswb`.
- `i16x8.narrow_i32x4_u` now lowers directly to `VPpackusdw`.

That preserves Wasm's signed-source interpretation for the input lanes, so negative lanes saturate to zero. Validation: `go test ./...` passes after pushing.

src/core/compiler/backend/railshot/simd_exec_test.go
jtenner: Fixed in https://github.com/wago-org/wago/commit/3d9f2cfc702f371deaa42191617b33a77d5e35cb.

What changed:
- Updated the `i8x16.narrow_i16x8_u` expected vector so negative signed i16 inputs saturate to `0`.
- Updated the `i16x8.narrow_i32x4_u` expected vector so negative signed i32 inputs saturate to `0`.
- The same test case now explicitly covers those negative-input lanes that distinguish correct `PACKUS*` behavior from unsigned pre-clamping.

Validation: `go test ./...` passes after pushing.

src/core/compiler/backend/railshot/simd.go
jtenner: Addressed in https://github.com/wago-org/wago/commit/3d9f2cfc702f371deaa42191617b33a77d5e35cb.

What changed:
- Replaced the scalar per-lane extract/min/max/insert lowering with the branchless packed vector sequence.
- Added the needed immediate shift helpers (`VPsrldImm`, `VPsrlqImm`) for NaN canonicalization.
- Existing packed float min/max tests cover NaN and signed-zero behavior.

Validation: `go test ./...` passes after pushing.

PR #210, opened 2026-07-08
wago: enable multi-value support

jtenner:
## Summary
- split multi-value support from `wasm2` onto updated `main`
- enable `CoreFeatureMultiValue` in the supported/default feature set while keeping reference-types and tail-call rejected
- add decoder/validator coverage for multi-result block type indexes and branch payloads
- add runtime/API coverage for multi-result if, br, br_if, br_table, typed `Call`, raw `Invoke`, compiled-blob roundtrip, and cross-instance imported calls
- update `FEATURES.md` and example output for the new default feature set

## Tests
- `go test ./src/core/compiler/wasm -run TestMultiValueBlockTypeDecodeAndValidation -count=1`
- `go test ./src/core/compiler/frontend -count=1`
- `go test ./src/wago -run 'TestMultiValue|TestCrossInstanceCallMultiValueImport' -count=1`\n- `go test ./...` fails on updated `main` in `cli/wago`: `registrySearch`, `registryInfo`, `registryVersions`, `registryStar`, `registryUnstar`, and `registryToken` are undefined.


jtenner:
@JairusSW could you please review the updated multivalue test coverage?

JairusSW (approved):
Good

PR #231, opened 2026-07-10
Complete WebAssembly 2.0 runtime support

jtenner:
## Summary

- close the WebAssembly 2.0 decode, validation, compilation, execution, host-boundary, and product-surface gaps while preserving strict malformed-module rejection
- implement nullable and non-null `funcref`, store-bound `externref`, reference globals, typed element segments, multiple local/imported tables, indexed table operations, and exact reference ownership across instances
- add bounded public owners for host funcrefs, externrefs, reference globals, and externref tables, including close-order and start-trap lifetime guarantees
- add structural-only codec v20 metadata for reference globals, tables, elements, limits, and required feature families without serializing live host/store identity
- make snapshots fail closed for every table/reference-global module and make `Class` pooling fresh-reinstantiate all reference-bearing state
- expose deterministic Wasm-index metadata for every function, global, and table after compile or codec load
- optimize `ResetMemorySnapshot` for measured-faster explicit-bounds zero/one-page instances while retaining reinstantiation for larger or unsupported shapes
- pin and execute the official WebAssembly Release 2 corpus with zero module, assertion, and product-gap skips

## Correctness and safety boundaries

- malformed structured custom sections remain hard decode errors
- multiple memories remain explicitly rejected
- table64, tail calls, threads/atomics, and broader post-Release-2 proposals remain out of scope
- raw or unowned non-null reference descriptors remain fail closed at public boundaries
- mmap-backed runtime state contains no Go pointers
- reference tokens/refSlots are validated before use
- table-0 direct execution paths and documented runtime layouts remain unchanged
- snapshots reject tables and reference globals until a resolver/live-state format exists
- imported reference globals/tables are rejected by pooled `Class` reset because their host-owned state cannot be tenant-reset safely

## Measured Class reset optimization

`BenchmarkClassAcquireRelease`, AMD Ryzen 7 8845HS, explicit bounds, 500 ms runs:

| Fixture | Reinstantiate median | Snapshot median | Delta |
|---|---:|---:|---:|
| globals / zero memory | ~1,295 ns | ~1,016 ns | **-21.5%** |
| one memory page | ~1,260 ns | ~1,023 ns | **-18.8%** |
| allocation | 1,312 B / 11 allocs | 25 B / 2 allocs | **-9 allocs** |

Unconditional copying lost at two pages and worsened with size, so the implementation deliberately falls back above one Wasm page. The benchmark retains 2/4/8/16/64-page cases to guard the crossover.

## Validation

- focused reference, table, codec, snapshot, pool, Class, metadata, and close-order tests under `go test -race`
- `go test ./... -count=1`
- `go test -race ./... -count=1`
- codec generated-valid and mutation fuzzers, 5 seconds each
- `make spec1`: 629 modules, 16,026 assertions, zero failures/skips
- Release 2 validation: 1,600 modules, 2,880 assertions, zero failures
- `make spec2`: 1,600 modules, 48,248 assertions, every tracked gap zero
- `scripts/spec-card_test.sh`
- `go generate ./`
- `go test -tags wago_guardpage ./src/wago -run '^TestClass' -count=1`
- `git diff --check`

## Documentation

Updated `FEATURES.md`, `ROADMAP.md`, `README.md`, `agent-todo.md`, `docs/reference-values.md`, and `docs/spec-testing.md` with the completed support matrix, product boundaries, ownership rules, spec workflow, layout watchpoints, and measured reset crossover.


jtenner:
@codex review

JairusSW (approved):
phat pr

src/core/compiler/backend/railshot/table.go
```diff
@@ -76,20 +103,25 @@ func (f *fn) tableInit(r *wasm.Reader) error {
 	f.a.Load64(RSI, RSP, f.spillOff(d-2)) // src element offset
 	f.a.Load64(RCX, RSP, f.spillOff(d-1)) // n entries
 
-	f.a.Load64(R8, RBX, -int32(offTablePtr))
+	f.loadTableDescriptor(R8, tableIdx)
 	f.a.Load32(RAX, R8, 0)
 	f.a.LeaScaled(RDX, RDI, RCX, 0, 0)
 	f.trapUnlessLE(RDX, RAX)
-	f.tableEntryAddr(RDI, R8)
+	externref := f.elementIsExternref(elemIdx)
```
JairusSW: **Correctness — OOB heap write.** `table.init` derives the copy stride and byte count from the *element segment* classification (`elementIsExternref`, which matches only exact `externref`), while the destination address uses the *table* type (`typedTableEntryAddr` → `tableIsExternref`). Validation permits `noextern <: externref` (subtyping, see `validate_ops.go` `refSubtype`), so a `(ref null noextern)` passive segment into an `externref` table makes `externref == false` here → the `rep movsb` copies `n*32` bytes into an 8-byte-per-entry table = out-of-bounds heap corruption.

The destination entry size is fixed by the table type, so all three strides (dest addr, source addr, byte count) should be keyed on `f.tableIsExternref(tableIdx)`, not the segment. Fixed in follow-up.

src/core/compiler/backend/railshot/table.go
```diff
@@ -11,32 +9,32 @@ import (
...
-func readSingleTableIndex(r *wasm.Reader, op string) error {
-	idx, err := r.U32()
-	if err != nil {
-		return err
-	}
-	if idx != 0 {
-		return fmt.Errorf("%s: multi-table unsupported: table %d", op, idx)
-	}
-	return nil
+func readSingleTableIndex(r *wasm.Reader, _ string) (uint32, error) {
```
JairusSW: **Cleanup — dead parameter.** The `op string` parameter (now `_ string`) is unused after the multi-table error strings were removed; every caller passes a string literal (`"table.size"`, `"table.copy"`, …) that is immediately discarded. Drop the parameter and the literal arguments in all callers. Fixed in follow-up.

src/core/compiler/backend/railshot/table.go
```diff
@@ -76,20 +103,25 @@ func (f *fn) tableInit(r *wasm.Reader) error {
...
 	disp := int32(elemIdx) * runtime.PassiveElemDescBytes
 	f.a.Load64(R8, RBX, -int32(offPassiveElemPtr))
 	f.a.Load32(RAX, R8, disp+8)
 	f.a.LeaScaled(RDX, RSI, RCX, 0, 0)
 	f.trapUnlessLE(RDX, RAX)
 	f.a.Load64(R8, R8, disp)
-	f.entryArrayAddr(RSI, R8)
-	f.a.ShiftImm(4, RCX, 5, true)
+	f.entryArrayAddr(RSI, R8, externref)
+	shift := byte(5)
```
JairusSW: **Cleanup — duplicated stride idiom.** The `shift := byte(5); if externref { shift = 3 }` pattern (8-byte externref vs 32-byte funcref entry stride) is repeated in `entryArrayAddr`, `tableInit`, and `tableCopy`. Extract a single `entryStrideShift(externref)` helper so the stride constant lives in one place and cannot drift if `TableEntryBytes` changes. Fixed in follow-up.

src/wago/valtype.go
```diff
@@ -58,22 +84,36 @@ func valTypesFromWasm(ts []wasm.ValType) []ValType {
 	return out
 }
 
-// code is the wasm value-type byte (used by the codec).
-func (t ValType) code() byte {
+// code is the wasm value-type byte used by the current compiled-module codec.
+// Codec version 18 defines the reference type codes as structural metadata;
```
JairusSW: **Doc fix.** Comment says "Codec version 18 defines the reference type codes", but the actual codec version (`wagoVersion` in `api.go`) is `20`. Fixed in follow-up.

src/wago/instantiate.go
```diff
@@ -756,6 +998,324 @@ func (in *Instance) resetToSnapshot(s *Snapshot) error {
 // host-imported one). Use Memory().Bytes() for the zero-copy byte view.
 func (in *Instance) Memory() *Memory { return in.memory }
 
+type hostFuncRefAttachments struct {
```
JairusSW: **Cleanup — triplicated dedup helpers (not auto-applied).** `hostFuncRefAttachments`, `globalImportAttachments`, and `tableImportAttachments` (plus `detachImportedHostFuncRefs/Globals/Tables`) triplicate the same inline-array + overflow-slice dedup/attach/detach logic — ~250 lines differing only in element type and method names. A single generic `[T comparable]` dedup set would collapse all three, so a future fix to the overflow handling need only be made once. Left as-is in this pass: large structural refactor of import-lifecycle code, out of scope for an automated fix.

src/core/compiler/frontend/frontend.go
```diff
@@ -950,6 +1192,24 @@ func (p supportPass) constExprBytes(body []byte, context string) error {
...
+	case 0xd0:
+		heap, err := r.S33()
+		if err != nil {
+			return err
+		}
+		if !p.feat.ReferenceTypes {
+			return p.unsupported("const expression", "ref.null (reference-types disabled)", instructionContext(context, 0))
+		}
+		if heap != -16 && heap != -17 {
+			return p.unsupported("const expression", fmt.Sprintf("ref.null heap type %d", heap), instructionContext(context, 0))
```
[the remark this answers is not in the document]
JairusSW: Fixed in 3de494a + 9cae6f4. The func/extern-only `ref.null` whitelist turned out to be replicated in **three** decode sites, not just this one — the support pass here, `constexpr.go` (global initializers, the path a `funcref` global with `ref.null nofunc` actually hits first), and `element_expr.go` (element-segment exprs). All three now accept `nofunc` (-13) / `noextern` (-14) and map them to their func/extern-family null. Added hand-assembled regression fixtures (the wabt on our hosts cannot emit these heap types) covering globals and active elements; verified passing on linux/amd64.

src/wago/instantiate.go
JairusSW: Done in 3de494a — collapsed all three attachment sets and the three `detachImportedX` dedup loops onto a single generic `importDedup[T comparable]` (~200 lines removed), preserving behavior including tables validating on every import occurrence before the dedup check. Full suite green on linux/amd64.

PR #350, opened 2026-08-05
amd64: emit compact EVEX tuple displacements

### LINKED ISSUE #348: [P2][AMD64] Emit compact EVEX tuple-scaled memory displacements
## Context

Discovered while implementing PR #347, which adds compact ordinary ModRM/SIB displacement selection for issue #324.

The AVX-512 plugin encoder in `src/core/encoder/amd64/asm_avx512.go` remains separate intentionally. `evexMemIdx` currently always emits `mod=10 + disp32`, including zero and small displacements, for `ZMovdqu64LoadIdx` and `ZMovdqu64StoreIdx`.

EVEX cannot safely reuse the ordinary scalar/SSE/VEX signed-byte selector. EVEX `disp8` is compressed and tuple-scaled: for a full 512-bit memory tuple, the encoded signed byte is multiplied by 64. For example, a byte displacement of 64 must be encoded as compressed `disp8=1`, not `disp8=64`.

PR #347 deliberately leaves EVEX unchanged rather than emitting incorrect addresses.

## Scope

- [ ] Add an EVEX-specific displacement selector parameterized by the instruction's tuple scale `N`.
- [ ] Use `mod=00` with no displacement when the byte displacement is zero and the base permits it.
- [ ] Preserve RBP/R13 zero displacement with `mod=01` and compressed `disp8=0`.
- [ ] Use `mod=01` only when `disp % N == 0` and `disp/N` fits signed 8-bit range `[-128, 127]`.
- [ ] Emit the signed quotient `disp/N` as the compressed `disp8`.
- [ ] Fall back to `mod=10 + disp32` for unaligned displacements or quotients outside the signed-byte range.
- [ ] Keep RSP/R12 SIB handling and extended base/index/register bits unchanged.
- [ ] Model tuple scale explicitly so future EVEX instructions can select their correct full/half/scalar/broadcast tuple semantics rather than assuming 64 bytes globally.
- [ ] Do not route EVEX operands through the ordinary byte-displacement helper added by PR #347.

## Suggested implementation

Introduce a helper along these lines:

```go
func evexAddrMode(base Reg, disp, tupleScale int32) (mod byte, encodedDisp int32) {
    if disp == 0 && base&7 != 5 {
        return 0x00, 0
    }
    if tupleScale > 0 && disp%tupleScale == 0 {
        q := disp / tupleScale
        if q >= -128 && q <= 127 {
            return 0x40, q
        }
    }
    return 0x80, disp
}
```

Have `evexMemIdx` accept or derive the tuple scale, emit the existing ModRM/SIB bytes with the selected mode, then emit either no displacement, the compressed quotient byte, or the original byte `disp32`.

For the current `vmovdqu64 zmm, m512` load/store helpers, the full-memory tuple scale is 64 bytes.

## Correctness tests

For tuple scale 64, add exact byte goldens around compressed-displacement boundaries:

- [ ] `-8,256` bytes (`-129 * 64`) → `disp32`
- [ ] `-8,192` bytes (`-128 * 64`) → compressed `disp8=-128`
- [ ] `-64`, `0`, `64`, and `8,128` bytes (`127 * 64`) → compact forms
- [ ] `8,192` bytes (`128 * 64`) → `disp32`
- [ ] Unaligned byte displacements such as `1`, `63`, `65`, and `127` → `disp32`, never compressed
- [ ] RSP, R12, RBP, and R13 base behavior
- [ ] Extended EVEX register, base, and index bits
- [ ] Both `ZMovdqu64LoadIdx` and `ZMovdqu64StoreIdx`
- [ ] Cross-check generated bytes with an independent AVX-512-capable disassembler or assembler

## Measurement

- Report bytes removed from representative wide-SIMD plugin fixtures.
- Count no-displacement, compressed-disp8, and disp32 selections.
- Verify no per-instruction allocation is introduced.
- Confirm runtime behavior on an AVX-512-capable host where available; byte-level encoding tests remain mandatory on all hosts.

## Acceptance criteria

- Every current EVEX memory operand uses the shortest correct encoding for its exact tuple type.
- Unaligned byte displacements never use an incorrect compressed `disp8`.
- RBP/R13 and RSP/R12 special forms remain correct.
- No generated fixture becomes larger.
- PR #347 remains focused on ordinary scalar/SSE/VEX displacement semantics; this issue owns EVEX tuple-scaled completion.

jtenner:
## Summary

- select no-displacement, compressed-disp8, or disp32 EVEX memory forms
- model the EVEX tuple scale explicitly
- add exact boundary and special-base encoding coverage

## Correctness

- compressed disp8 is emitted only for tuple-aligned displacements
- the encoded byte is disp/tupleScale
- unaligned and out-of-range displacements retain their original disp32 form
- RBP/R13 and RSP/R12 addressing behavior is preserved

## Tests

- `go test ./src/core/encoder/amd64 -count=1` — passed
- `go test ./... -count=1` — blocked by the test environment: the installed WABT cannot parse the pinned Release 3 GC, exception-handling, and memory64 fixtures, and `WAGO_SPEC_INTERPRETER` is unset (the tests require the pinned official Release 3 interpreter)
- `git diff --check` — passed
- independent assembler/disassembler check: GNU as and objdump 2.44 — verified `vmovdqu64` indexed forms for displacements 64 (`disp8=01`), -8192 (`disp8=80`), 65 (`disp32=41 00 00 00`), RBP zero (`disp8=00`), R12/R13 addressing, and the 64-byte store form

## Risk

Low and isolated to the AVX-512 indexed-memory encoder. No public API, IR, validator, GC, or non-EVEX encoding changes.

Fixes #348


JairusSW:
good

JairusSW:
ready for review?

PR #361, opened 2026-08-07
threads: add bounded concurrent atomics product

JairusSW:
## What changed

- adds the opt-in `CoreFeatureThreads` product for Linux and macOS on amd64/arm64 with explicit bounds
- gives eligible shared-memory instances private basedata/control mappings and instance-local native execution leases, allowing distinct instances to execute concurrently while ordinary modules retain the process-wide lease
- lowers the complete classic scalar atomic matrix directly in AMD64 and ARM64 Railshot: load/store, fence, add/sub/and/or/xor/exchange RMWs, and compare-exchange across every width
- adds bounded `wait32`, `wait64`, and `notify` state keyed by exact shared-memory identity and offset, with timeout, cancellation, instance-close, and memory-close handling
- persists exact Threads and wait-helper requirements in artifact codec v32 and exposes them through module metadata, the public facade, CLI schema, feature ledger, roadmap, and architecture docs
- executes the checked-in official atomic proposal core and adds overlap, contention, lifecycle, trap-integrity, race, allocation, footprint, and benchmark coverage

## Product boundary

This first product intentionally requires one imported shared memory32 with an exact maximum and explicit bounds. It admits numeric functions/globals and classic atomics. It rejects memory growth, memory64, multi-memory, signal bounds, function imports, tables, tags, segments, WasmGC, exceptions, snapshots, shared-everything GC, and concurrent entry into the same instance before native code generation.

Threads remain experimental and are not included in `CoreFeaturesV3`.

## Why

Wago already had shared memory, but native activations were guarded by one process-wide lease because mutable execution control lived beside the backing memory. Lowering atomics alone would therefore have produced correct isolated results without concurrent native execution.

This change separates private activation control from shared linear-memory backing for the bounded product, preserving ordinary generated code while enabling actual parallel execution across distinct instances.

## Validation

- `go test ./... -count=1 -timeout=180s`
- `go test -race ./src/wago -run '^TestThreads' -count=1 -timeout=180s`
- all Threads tests pass under native Darwin/arm64 and Rosetta Darwin/amd64
- `src/wago` test binaries cross-compile for linux, darwin, and windows on amd64 and arm64
- official atomic core: 63 setup actions, 149 exact returns, 45 exact traps
- official wait/notify core: 3 exact returns, 6 exact bounds/alignment traps
- warmed direct atomic execution: 0 allocations

## Performance and footprint

Apple M4 Max, Darwin/arm64:

- direct atomic load/store/RMW/CAS: about 88–95 ns/op, 0 allocations
- fixed-work throughput versus the serialized ordinary path:
  - 2 workers: 1.90x, +90.4%
  - 4 workers: 3.68x, +267.6%
  - 8 workers: 7.30x, +629.7%
- `Compiled`, `Instance`, and `Memory` Go object sizes are unchanged
- each threaded instance owns a private 65,824-byte logical control mapping
- compile and scalar instantiate deltas are within noise
- known cost: ordinary `Invoke` is about 12% slower and ordinary shared-memory public entry is about 9–12% slower in local A/B runs; both remain zero-allocation

The complete implementation boundary and raw benchmark record are in `docs/threads-atomics-plan.md`.


src/core/compiler/backend/railshot/amd64/atomic.go
```diff
@@ -0,0 +1,220 @@
...
+	expected = relocateRAX(expected)
+	f.spillIfUsed(RAX)
+	f.pinned = f.pinned.add(RAX)
+	if d.ResultSize == 8 {
+		f.a.MovReg64(RAX, expected)
+	} else {
+		f.a.MovRegReg32(RAX, expected)
+	}
+	base, ea, disp := f.atomicMem(d.Offset, int(d.Size))
+	f.a.LockCmpxchgIdx(base, ea, replacement, disp, int(d.Size))
```
jtenner, line 83: [P1] Zero-extend successful i64.atomic.rmw32.cmpxchg_u results. A successful 32-bit CMPXCHG does not write EAX, so RAX retains the full 64-bit expected operand loaded above. This returns upper bits that Wasm requires to be zero. The official corpus reproduces it: got 0x1111111111111111, want 0x11111111. The size-4/result-size-8 case needs explicit normalization just like the byte and halfword cases.

src/wago/instance_native_context.go
```diff
@@ -20,20 +20,29 @@ var (
...
+type executionLease struct{ local *sync.Mutex }
 
 // beginNativeEntry acquires the serialized execution lease and rebinds this
 // instance's pointer context before native code can observe basedata. Memory
 // size/growth fields remain backing-owned; invocation control is refreshed by
 // the engine entry/resume paths.
-func (in *Instance) beginNativeEntry() (*executionLease, error) {
+func (in *Instance) beginNativeEntry() (executionLease, error) {
+	if in.c.threadedMemory0() {
+		mu := &in.memoryDir.nativeMu
```
jtenner, line 31: [P1] This per-instance mutex does not protect all of the mutable instance state that the old process-wide lease protected. Invoke marshals into in.serArgs and later reads in.results before/after beginNativeEntry, so two calls into the same threaded instance can race and one call can execute with the other call’s arguments/results. The documented product boundary says simultaneous calls into one instance are rejected, but beginInvocation permits them and this mutex only serializes native entry after scratch has already been touched. Separately, GlobalValue/SetGlobalValue still take nativeExecutionMu, not this mutex, so admitted numeric globals can race with guest execution. The instance-level exclusion must cover the whole public call boundary and host access to native-visible instance state (or concurrent entry must fail before any scratch mutation).

src/wago/instantiate.go
```diff
@@ -372,7 +373,18 @@ func (b *instanceBuilder) instantiate() (result *Instance, err error) {
 			runtime.ReleaseEngine(eng)
 			return nil, fmt.Errorf("imported memory %q: %w", c.memoryImport, err)
 		}
-		jm, memObj = m.jobMemory(), m
+		memObj = m
+		if c.threadedMemory0() {
```
jtenner, line 377: [P1] Verify that the imported Memory is actually shared before selecting the threaded path. validateLimits checks address width and limits but not memoryStateShared, and attachImporter accepts the first importer of NewMemory. Consequently a module declaring a shared import can instantiate with ordinary NewMemory and execute atomics against a provider whose external type is non-shared, violating Wasm import type matching and the concurrency contract.

src/core/compiler/frontend/frontend.go
```diff
@@ -1459,10 +1465,17 @@ func (p supportPass) instrByte(r *wasm.Reader, op byte, context string, instr in
...
+		if err != nil {
 			return false, err
 		}
-		return false, p.unsupported("instruction", fmt.Sprintf("opcode 0x%02x", op), ctx())
+		if !wasm.IsCoreAtomicInstructionKind(imm.Kind) {
+			return false, p.unsupported("instruction", imm.Kind.String(), ctx())
+		}
+		if !p.feat.Threads {
+			return false, p.unsupported("instruction", imm.Kind.String()+" (threads disabled)", ctx())
+		}
```
jtenner, line 1477: [P2] Enabling this family makes the validator’s permissive atomic alignment check observable. checkSharedMemArg delegates to checkMemArg, which only rejects ma.Align > natural; the threads proposal requires atomic memargs to use exactly their natural alignment. Modules encoding a smaller alignment are therefore accepted as valid and compiled. Add an equality check for atomic operations (including wait/notify) rather than reusing the ordinary load/store <= rule.

jtenner (changes requested):
The implementation has several correctness gaps beyond the current CI failures. The inline comments cover the merge-blocking cases I found. In particular, the new per-instance execution lease is acquired too late to protect invocation scratch and is not shared by host global access; imported memory sharedness is not checked; AMD64 narrow cmpxchg returns a non-canonical result; and enabling atomics exposes an existing validation rule that accepts under-aligned atomic memargs.

jtenner (commented):
One additional aliasing case from the continued pass:

src/wago/instance_native_context.go
jtenner, line 31: [P1] The local lease also fails to serialize mutable imported globals shared by distinct threaded instances. GlobalImport explicitly preserves one Global object across instances, and each instance’s native global cell points at that shared storage. Two instances using the same mutable numeric global now hold different memoryDir.nativeMu locks and can concurrently issue ordinary non-atomic loads/stores to the same cell. Either exclude mutable global imports from this product, give shared globals a common execution lease, or otherwise make that aliasing safe.

src/core/compiler/backend/railshot/amd64/atomic.go
JairusSW: Proven. On a successful 32-bit CMPXCHG, EAX/RAX is not rewritten, so the full 64-bit expected operand can escape. e1377733 explicitly normalizes RAX through a 32-bit self-move and strengthens the i64_32 regression to use 0x1111111111111111 while expecting 0x11111111. The AMD64 regression passes under Rosetta, and the full suite is green.

src/wago/instance_native_context.go
JairusSW: Proven. The red race test showed concurrent Invoke calls touching reusable instance scratch before the old native-entry lock; a deterministic guest/host test also showed SetGlobal completing during guest execution. e1377733 adds a threaded-only invoke mutex covering the entire public Invoke boundary and keeps the native activation mutex shared with every host global accessor. Both regressions pass under go test -race; direct calls remain at 0 allocs/op.

src/wago/instantiate.go
JairusSW: Proven. A focused regression instantiated a declared shared-memory module with NewMemory and failed because it was accepted. e1377733 records Wasm sharedness separately from host multi-import capability and rejects a shared import backed by a non-shared provider. The regression and existing ordinary/shared import suites pass.

src/core/compiler/frontend/frontend.go
JairusSW: Proven. A module using i32.atomic.rmw.add with align=2 bytes instead of its natural 4-byte alignment validated before the fix. e1377733 requires exact natural alignment in checkSharedMemArg, covering loads/stores, RMW, cmpxchg, wait, and notify; validator fixtures now encode valid natural alignments explicitly. Compiler and full repository suites pass.

src/wago/instance_native_context.go
JairusSW: Proven. The focused regression showed that a threaded module admitting a mutable imported global could create a shared non-atomic cell governed by distinct per-instance leases. e1377733 rejects mutable global imports at the threaded execution boundary and documents that boundary. The rejection regression and full suite pass.

PR #364, opened 2026-08-07
gc: add shadow tracing and hardened stress verification

### LINKED ISSUE #301: [P0][GC] Add shadow tracing, forced-collection stress, and transactional failure injection
## Context

Every collector optimization must preserve stale-reference rejection, exact roots, trap order, barriers, subtype semantics, and transactional allocation. Build a reusable correctness oracle before substantially changing minor collection, cards, promotion, metadata, or old-space layout.

### Stress collection

- [ ] Support collection on every allocation in dedicated stress builds.
- [ ] Randomly choose minor or full collection at admitted allocation sites.
- [ ] Randomly collect at native-call safepoints and host-transition boundaries.
- [ ] Run the matrix under Throughput, Tiny, explicit bounds, guard-page mode, amd64, and arm64.

### Shadow verification

- [ ] Implement a slow non-moving/reference tracing oracle for tests.
- [ ] Recompute the complete old-to-young edge set and compare it with the remembered/card structure.
- [ ] Verify exact root sets against conservative test-only frame/local enumeration where possible.
- [ ] Detect unnecessary root retention as well as missing roots.
- [ ] Verify public-token, global, table, cross-instance, exception, and snapshot temporary roots.

### Failure injection

- [ ] Inject failures at every promotion planning, destination allocation, commit, handle publication, card growth, and backing-growth point.
- [ ] Prove failed transactional promotion leaves handles, objects, roots, nursery state, and old-space metadata unchanged.
- [ ] Inject handle-batch cancellation and epoch changes around collection and `Close`.
- [ ] Poison evacuated nursery bytes and freed old/tiny spans in stress builds.

### Fuzzing and semantic gates

- [ ] Fuzz object sizes, alignments, field layouts, rec groups, packed fields, nullability, mutability, subtyping, and maximum array lengths.
- [ ] Fuzz roots through recursion, direct/indirect calls, `call_ref`, tail calls, EH, re-entry, and cross-instance call chains.
- [ ] Test old-to-young writes in every collector phase.
- [ ] Test sparse/dense cards, card overflow, metadata growth, and giant reference arrays.
- [ ] Preserve effect and trap order for dead-allocation elimination, cast removal, load forwarding, bounds elimination, and bulk lowering.
- [ ] Keep unsupported collection-disabled products explicitly fail-closed on exhaustion.

## Acceptance criteria

- The stress suite can be run independently and in CI-selected shards.
- Every optimized collector path has a shadow/oracle comparison.
- Failure injection demonstrates atomicity rather than merely checking returned errors.
- Hardened builds retain expensive internal checks without imposing them on release hot paths.

jtenner:
Closes #301.

## What changed

- add an independent descriptor-driven shadow tracer for Throughput and Tiny GC
- compare full-collection liveness against exact roots, globals, tables, and object graphs
- cross-check mutable and direct native root enumeration exactly
- make failed promotion planning restore nursery, handles, marks, cards, remembered metadata, and throughput allocator state
- add collector-scoped `wagodebug` failure injection for promotion planning, destination allocation, commit preflight, handle publication, card growth, and backing growth
- randomize minor/full collection at allocation safepoints in hardened builds
- integrate the shadow oracle into randomized and fuzz stress tests
- add native handle-batch cancellation and epoch checks around successful and failed collection plus `Close`
- add native amd64/arm64 explicit/guard-page hardening PR and scheduled CI shards
- rebase onto current `main` and make the Tiny publication failure test prove allocator reuse behavior instead of depending on TinyGC's internal span representation

## Why

Collector optimizations need an oracle that does not share the production scanner and deterministic failures that prove atomicity rather than only checking returned errors. The previous promotion rollback freed planned destinations, but that did not restore bump, backing, free-list, cached-span, or mark metadata exactly.

The behavioral TinyGC failure test also keeps this hardening useful across allocator metadata changes: after injected publication failure, handles remain unchanged and the only available allocation remains reusable.

## Release impact

The expensive oracle remains test-only. Failure injection is build-tagged behind `wagodebug`; the normal test binary contains no injector symbols or injector state. Promotion rollback uses one transaction header and reverses successful free-span allocations in LIFO order, retaining the baseline 24-byte promotion plan instead of copying allocator metadata per survivor.

## Validation

- `go test ./src/core/runtime/gc -count=3`
- `go test -tags wagodebug ./src/core/runtime/gc -count=3`
- `go test -race -tags wagodebug ./src/core/runtime/gc -count=1`
- `go test -tags='wagodebug wagostress' ./src/core/runtime/gc`
- `go test -tags='wagodebug wagoguard' ./src/core/runtime/gc`
- shuffled/repeated fault isolation and rollback tests, including mixed class/large/bump sequences
- bounded fuzzing of Throughput operations, Tiny operations, and descriptor validation (more than 700,000 executions total)
- focused native host-transition and frame-root tests in explicit and guard-page builds
- native Linux amd64/arm64 explicit and guard-page CI hardening matrix
- `make lint`, `go vet ./src/core/runtime/gc`, and `actionlint`

After rebasing onto `main` at `5e6ebc21`, a 10-run Ryzen 7 7800X3D comparison measured `BenchmarkForcePromoteTransactional` at 115.27 ns/op on `main` versus 115.81 ns/op on this branch (+0.47%). Both report 0 B/op and 0 allocations/op. Promotion-plan footprint remains 24 bytes per survivor. Earlier constructor and remembered-array-write comparisons remained within 2.2% of baseline.

The repository-wide local suite requires network listener access and the external `tests/spec-v3` checkout; affected GC package suites pass, and normal CI provisions those requirements.

### DIFF
--- a/.github/workflows/regression-stress.yml
+++ b/.github/workflows/regression-stress.yml
@@ -4,6 +4,13 @@ on:
   schedule:
     - cron: "23 4 * * 1"
   workflow_dispatch:
+  pull_request:
+    branches: [main]
+    paths:
+      - '.github/workflows/regression-stress.yml'
+      - 'docs/gc.md'
+      - 'src/core/runtime/gc/**'
+      - 'src/wago/*gc*'
 
 permissions:
   contents: read
@@ -52,6 +59,7 @@ jobs:
         run: go test -count=3 -tags '${{ matrix.tags }}' -run 'Test(RuntimeGCHostFuncRefCallAndTailOwnership|GenericGCResultsIssueBoundedHostTokens|GCCrossInstanceCallsWithSharedPersistentRoots|GCHostReentryNativeFrameRoots|GCArm64HostReentryRoots|GCWarmSnapshotPreservesCyclesAndSharing)$' ./src/wago
 
   stress:
+    if: github.event_name != 'pull_request'
     runs-on: ubuntu-24.04
     timeout-minutes: 45
     steps:
--- a/.github/workflows/regression-stress.yml
+++ b/.github/workflows/regression-stress.yml
@@ -16,6 +16,41 @@ env:
   GO_VERSION: "1.22"
 
 jobs:
+  gc-hardening:
+    name: GC hardening / ${{ matrix.arch }} / ${{ matrix.bounds }}
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - arch: amd64
+            runner: ubuntu-24.04
+            bounds: explicit
+            tags: wagodebug
+          - arch: amd64
+            runner: ubuntu-24.04
+            bounds: guard-page
+            tags: wagodebug,wago_guardpage
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+            bounds: explicit
+            tags: wagodebug
+          - arch: arm64
+            runner: ubuntu-24.04-arm
+            bounds: guard-page
+            tags: wagodebug,wago_guardpage
+    runs-on: ${{ matrix.runner }}
+    timeout-minutes: 30
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4
+      - uses: actions/setup-go@40f1582b2485089dde7abd97c1529aa768e1baff # v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+          cache: false
+      - name: Run collector oracle, injection, fuzz seeds, and stress suites
+        run: go test -count=3 -tags '${{ matrix.tags }}' ./src/core/runtime/gc
+      - name: Run native GC root and host-transition stress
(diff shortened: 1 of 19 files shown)


JairusSW:
Ready-for-review update:

- rebased onto current `main` (`5e6ebc21`)
- adapted the TinyGC publication-failure test to verify allocator reuse rather than private span metadata
- restored merge-result formatting and pushed compatibility commit `08004b5e`
- reran the normal, `wagodebug`, `wagostress`, and `wagoguard` GC suites successfully
- all four amd64/arm64 explicit/guard-page GC-hardening CI jobs are green
- refreshed 10-run benchmark: 115.27 ns/op on `main` vs 115.81 ns/op on this PR (+0.47%), with 0 B/op and 0 allocs/op on both

The earlier integration failures were compile/format fallout from TinyGC metadata changes merged after this branch was created; no runtime correctness regression was found.

PR #365, opened 2026-08-07
compiler: precompute Wasm GC layouts

### LINKED ISSUE #306: [P1][Compiler] Precompute canonical GC type, layout, scan, and barrier metadata
## Context

Current AMD64 GC lowering repeatedly flattens recursive type groups, walks preceding struct fields, recomputes offsets/layouts, classifies reference storage, and allocates per-site constructor-layout slices. These facts are immutable for a compiled module and should be computed once.

### Canonical module metadata

- [ ] Build a flat `typeIndex -> subtype` table during module preparation.
- [ ] Store direct rec-group base/length and local-to-canonical type mappings.
- [ ] Precompute every struct field's offset, width, alignment, mutability, storage kind, reference class, and nullability.
- [ ] Precompute array element layout and reference classification.
- [ ] Precompute final-struct object size/alignment, pointer-free status, and constructor slot mapping.
- [ ] Precompute compact reference scan programs/bitmaps for each canonical heap type.
- [ ] Precompute direct-native eligibility for get/set/cast/length/allocation operations.
- [ ] Precompute the initial write-barrier class for each mutable reference field/array type.
- [ ] Precompute subtype metadata needed by final and non-final checks.

### Compiler integration

- [ ] Replace per-site recursive-group walks and field scans with O(1) table lookups.
- [ ] Replace per-site `make` of constructor field layouts with immutable module-owned records.
- [ ] Intern common helper signatures and constructor layout IDs.
- [ ] Pass compact layout/type IDs to native stubs rather than re-deriving shapes.
- [ ] Persist and validate only the metadata required by serialized artifacts; reconstruct safe derived caches where preferable.

### Measurement

- [ ] Benchmark modules with many rec groups, many fields/type, thousands of casts, and thousands of allocation sites.
- [ ] Report compile time, B/op, allocs/op, retained metadata bytes, and emitted code bytes.
- [ ] Keep small scalar modules within noise.

## Correctness requirements

- Validate all source type declarations before publishing derived metadata.
- Preserve exact finality, recursive identity, declared subtyping, storage alignment, packed fields, mutability, and nullability.
- Reject overflow or malformed persisted metadata rather than trusting it.

## Acceptance criteria

- GC code generation performs no repeated linear field-layout walk for a known field.
- Native allocation eligibility performs no per-site heap allocation.
- Large WasmGC compile time and allocation traffic improve measurably without increasing generated code.

jtenner:
Closes #306.

## What changed

- build immutable compiler metadata alongside runtime GC type descriptors in one flattened type-section traversal
- precompute recursive-group bounds, struct field offsets/widths/alignment, constructor slots, scan offsets, storage/reference/nullability/mutability classes, initial barrier classes, array element layouts, object sizing, and direct/native eligibility
- pass the compiler-only metadata to AMD64 and ARM64 Railshot compilation
- replace repeated recursive-group and preceding-field walks with indexed lookups while retaining the legacy derivation path for low-level callers
- reuse immutable constructor plans in the production native allocation path without per-site allocation
- document that the side table points into validated decoded types and is discarded after code generation rather than serialized

## Why

GC type facts are immutable for a compiled module, but GC lowering repeatedly rediscovered them at instruction sites. Large modules with many recursive groups, wide structs, casts, and allocation/access sites consequently paid linear lookup and layout costs throughout code generation.

## Measurements

On an AMD Ryzen 7 7800X3D:

- worst-case lookup across 256 recursive groups and 64 fields: 597–611 ns/op legacy versus 15.17–15.26 ns/op precomputed, with 0 B/op and 0 allocs/op in both cases
- layout-heavy compilation with 2,000 constructor/last-field access sites: 5.57–6.38 ms/op legacy versus 4.83–5.37 ms/op precomputed
- emitted output is identical at 642,141 bytes
- allocation traffic for that compile workload is unchanged at approximately 8.20 MB and 2,087–2,088 allocations/op; native constructor-plan lookup itself is asserted at 0 allocations
- the synthetic 256-type × 64-field table retains 419,840 bytes of compiler-only layout metadata
- small scalar compilation remains at 79 allocs/op; timing samples overlap baseline noise

## Validation

- `go test ./src/core/...`
- targeted WasmGC integration tests in `./src/wago`
- AMD64 metadata/legacy equivalence and constructor slot-boundary tests
- ARM64 backend cross-build
- `go vet ./src/core/compiler/...`
- `git diff --check`

### DIFF
--- a/docs/gc.md
+++ b/docs/gc.md
@@ -410,7 +410,19 @@ Payload begins at `PayloadOffset == HeaderSize`, currently 16 bytes. Logical obj
 
 ## Compiled metadata and instantiation
 
-Frontend lowering produces immutable descriptor metadata during compile. `Compiled.GCTypeDescs` stores the descriptor slice so `.wago` blobs can instantiate without re-decoding the Wasm type section. The descriptor slice index matches flattened `wasm.TypeIdx.Index`, including function sentinels used only to preserve indexes. Codec v34 retains the appended `StorageV128` kind, the 16-byte layout contract, and validated native safepoint/callsite root maps; older codec versions are rejected.
+Frontend lowering produces immutable descriptor metadata during compile. The same
+flattened type-section traversal also builds a compiler-only table containing each
+type's recursive-group bounds, struct field offsets/sizes/initializer slots,
+array-element layout, alignment, object size, and collector-reference classification.
+Railshot indexes that table directly while compiling GC instructions instead of
+walking recursive groups and preceding fields at every use. The table points at the
+immutable decoded subtype declarations and is discarded after code generation; it is
+not retained by `Compiled` or serialized. `Compiled.GCTypeDescs` stores the runtime
+descriptor slice so `.wago` blobs can instantiate without re-decoding the Wasm type
+section. The descriptor slice index matches flattened `wasm.TypeIdx.Index`, including
+function sentinels used only to preserve indexes. Codec v34 retains the appended
+`StorageV128` kind, the 16-byte layout contract, and validated native
+safepoint/callsite root maps; older codec versions are rejected.
 
 Each `Instance` normally owns its own `gc.Collector` when its executable product can create or retain heap objects. Collectors are never shared across instances: nursery state, old-space state, roots, remembered sets, cards, and collection statistics are per-instance runtime state. MVP/non-GC modules keep `Instance.gc == nil` to avoid allocating an unused heap.
 
--- a/src/core/compiler/backend/railshot/amd64/compile.go
+++ b/src/core/compiler/backend/railshot/amd64/compile.go
@@ -183,10 +183,11 @@ const mergeFReg Reg = 11
 // fn holds the per-function code-generation state — the port's equivalent of
 // WARP's Compiler/backend working set. One is created per compiled function.
 type fn struct {
-	a  *amd64.Asm // the (reused) x86-64 encoder
-	s  *stack     // the valent-block operand stack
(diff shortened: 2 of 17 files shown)


jtenner:
Rebased onto `475967e4` (`main` as of August 10, 2026) and force-pushed as `70e1d0b3`.

Rebase notes:
- preserved the GC layout metadata path alongside #369's compact descriptors and #372's direct code-image assembly
- preserved the newer GC code telemetry sink in the compile options
- fixed the descriptor-builder error path introduced while resolving the compact-metadata overlap

Local validation:
- `go test ./src/core/...`
- targeted `./src/wago` GC struct/array/cast/native-allocation tests
- Linux/arm64 backend cross-compile
- `go vet ./src/core/compiler/...`
- `git diff --check`

Fresh Ryzen 7 8845HS measurements:
- field-layout lookup: 272–287 ns/op legacy vs 9.74–9.93 ns/op precomputed, 0 allocs/op
- 2,000-site layout-heavy compile: 4.08–4.30 ms/op legacy vs 3.59–3.85 ms/op precomputed
- output remains 642,141 code bytes; allocation count remains 2,085 allocs/op
- retained compiler-only layout metadata remains 419,840 bytes for the 256-type × 64-field fixture

The broad local `./src/wago -run 'GC|Struct|Array|Type'` selection also reached official-suite tests, but this checkout does not have `WAGO_SPEC_INTERPRETER` configured; CI has the pinned toolchain for those cells.

PR #374, opened 2026-08-10
gc: hoist native ABI checks and share resolver stubs

### LINKED ISSUE #307: [P1][JIT] Hoist collector ABI invariants and consolidate compact shared GC stubs
## Context

Current direct GC access sequences repeatedly validate immutable internal facts such as instance/collector ABI version, handle stride, and view structure. Complete checked sequences also consume substantial generated code. Semantic checks must remain, but internal invariants should be validated once and common work should live in compact shared stubs.

### Invariant validation

- [ ] Validate native collector ABI version, handle layout, type-map structure, and immutable view fields at artifact load and instantiation.
- [ ] Optionally retain one hardened native-entry guard in debug/hardened builds.
- [ ] Validate host-ingress references before publishing them into native Wasm state.
- [ ] Distinguish semantic checks from internal debug assertions.
- [ ] Preserve null traps, cast failure, bounds, mutability, ownership, and non-final subtype semantics.

### Shared stubs

- [ ] Move common final/non-final cast, test, typed access, array length, compact-reference resolution, and allocation slow paths into module- or runtime-level shared native text where practical.
- [ ] Use compact layout/type IDs produced by precomputed metadata.
- [ ] Keep slow/error paths in cold stub islands.
- [ ] Use short hot call-site sequences and relative calls.
- [ ] Separate collecting from noncollecting stubs in metadata.
- [ ] Define a leaf-stub ABI with an explicit clobber set that preserves hot Railshot registers.
- [ ] Avoid full `flush()` for noncollecting leaf stubs when exact liveness/clobbers permit.
- [ ] Deduplicate trap and helper stub families with one grouping/patching pass.

### Resolved-handle reuse

- [ ] Resolve a compact handle once across a safepoint-free valent-block region.
- [ ] Retain the compact handle as the exact root while using a transient raw object address.
- [ ] Invalidate the raw address on allocation, collection, host calls, cross-instance calls, backing relocation, or control merges that lose provenance.

## Measurement and gates

- [ ] Attribute JIT bytes and runtime to invariant checks, call-site setup, stub bodies, spills, and reloads.
- [ ] Require a meaningful code-size reduction on Dew and no more than a 2% unexplained runtime regression.
- [ ] Preserve stale/forged handle rejection at untrusted boundaries.
- [ ] Add equivalent correctness coverage on amd64 and arm64.

## Acceptance criteria

- Production hot access paths do not repeat immutable ABI-version/stride checks at every operation.
- Shared stubs reduce generated bytes without reintroducing Go transitions.
- Debug/hardened products can retain stronger internal validation.

jtenner:
Closes #307.

## Summary

- hoist immutable native-GC ABI checks to collector construction, artifact loading, and instance-view publication;
- record and fail closed on the native-GC ABI requirement in codec v35;
- keep every mutable semantic check dynamic: compact tag/range/liveness, current handle/space backing, object extent/type, array bounds, barriers/ownership, roots, lifecycle, EH/tails, snapshots, and host re-entry;
- add a `wagodebug` Go-to-native immutable-view assertion without charging production GC operations;
- replace repeated AMD64 direct scalar/length handle-resolution bodies with one 229-byte module leaf at the measured two-site crossover;
- retain local trap/source attribution and rel32 range checks;
- add one-entry raw-address reuse only while an unchanged compact local remains the root inside a mechanically safepoint-free straight-line region;
- invalidate reuse at calls, helpers/hosts, allocation, control/EH/tail edges, local writes, mutable-fact changes, and unknown/non-access GC opcodes;
- add ARM64 artifact/instantiation preflight coverage (ARM64 execution remains helper-lowered);
- add permanent code-size/runtime attribution, differential controls, tests, and docs.

## ABI and artifact boundary

`gc.ValidateNativeABI` checks the Go layouts consumed by native code, including handle entries, cards, allocation state, spaces, collector view, and instance view. `gc.BuildNativeInstanceView` validates collector identity, immutable type-map pointer/count, versions, and stride before basedata publication.

Generic struct/array artifacts now carry native ABI v6 in codec v35. Missing or mismatched requirements reject before execution. `compiledCodeCache` remains 64 bytes by packing the requirement with existing metadata markers.

## AMD64 codegen

The module resolver leaf:

- cannot allocate, collect, enter Go, publish roots, or become a safepoint;
- reloads mutable collector pointers/counts;
- validates handle range/liveness, space/backing extent, object extent, and canonical runtime type;
- returns only a resolved address plus a local trap class, leaving exact source trapping in each caller.

Controls:

- `WAGO_AMD64_NO_GC_SHARED_STUBS=1`
- `WAGO_AMD64_NO_GC_RESOLVE_REUSE=1`

## Measurements

Ryzen 7 8845HS, Go 1.24.4, linux/amd64:

| Fixture | Candidate | Differential | Delta |
|---|---:|---:|---:|
| one resolver site | 351 B inline-selected | 351 B inline | 0 |
| eight sites | 821 B shared | 1,669 B inline | -50.8% |
| 128 sites | 7,301 B shared | 24,349 B inline | -70.0% |
| eight-site reuse | 565 B, 1 resolve + 7 reuse | 821 B, 8 resolves | -31.2% |

Ten CPU-0-pinned 500 ms runtime samples:

- default shared + reuse: **310.0 ns/op median**;
- reuse disabled: **328.1 ns/op median**;
- shared + reuse disabled (fully inline): **324.0 ns/op median**;
- all: **0 B/op, 0 allocs/op**.

The default was 5.5% faster than shared-without-reuse and 4.3% faster than fully inline on the repeated-access fixture. Two isolated ~0.8 us host outliers did not affect medians.

A same-command stripped `wago_runtime` TinyGo build changed **2,096,928 -> 2,102,160 bytes** (+5,232, +0.250%). Collector, instance, native-view, and compiled-code-cache layouts do not grow.

## Qualification

Passed:

- `go test ./src/core/...`
- focused `src/wago` ABI, artifact, hot-path, resolver, telemetry, structural-corpus, and documentation tests
- `go test -race ./src/core/runtime/gc`
- focused `src/wago` race tests
- `-tags wagodebug`
- `-tags wago_gcstats`
- `-tags wago_guardpage`
- Linux ARM64 cross-build of Railshot and `src/wago`
- Darwin ARM64 cross-build of `src/wago`
- `git diff --check`

The complete `go test ./src/wago` run reaches no issue-specific failures, but remains red in this environment because the installed WABT cannot parse the pinned Core 3 fixtures and `WAGO_SPEC_INTERPRETER` is not configured. The failures are the existing official GC/EH/memory64/table64 fixture-conversion/accounting cases; focused corpus and changed-path tests pass.

## Commits

- `3b2daaf4` gc: trust validated native ABI boundaries
- `8e8657c6` docs: record issue 307 qualification

### DIFF
--- a/ARCHITECTURE.md
+++ b/ARCHITECTURE.md
@@ -13,15 +13,15 @@ on amd64 and arm64. Linux and Darwin/arm64 additionally support signal-backed
 guard-page bounds checks; all six targets support explicit bounds checks and
 cooperative cancellation safepoints.
 
-<!-- artifact:codec-version 34 -->
+<!-- artifact:codec-version 35 -->
 
-Compiled artifact v33 is a strict ordered section stream: a fixed header and
+Compiled artifact v35 is a strict ordered section stream: a fixed header and
 section count followed by length-delimited native-code and metadata sections.
 Unknown, duplicate, reordered, truncated, over-limit, and non-canonical section
 encodings fail closed. `Compiled.WriteTo` streams code without constructing a
 second full image; `Compiled.ReadFromWithLimits` reads code directly into an RW
 mapping, bounds code and metadata independently, validates all metadata, and
-seals that same mapping RX on first use. Version 32 and older artifacts are
+seals that same mapping RX on first use. Version 34 and older artifacts are
 intentionally rejected; Wago was unreleased, so there is no compatibility
 decoder or dual-format ambiguity.
 
@@ -68,7 +68,7 @@ callsite; amd64 adds hidden operand spill offsets, compact safepoint IDs, frame
 size, adapter return, and recursive call return-PC maps. The synchronous helper
 control frame publishes parked RSP, and Go exposes validated off-heap slots from
 each walked frame directly as mutable collector roots. Throughput/Tiny stress
-collection and the root walker remain zero-allocation after warm-up. Codec v34
+collection and the root walker remain zero-allocation after warm-up. Codec v35
 persists and strictly revalidates the map, including dynamic-import stack
 adjustments. Direct tail calls discard their caller frame. Numeric host callbacks
 use a bounded suspended-activation stack plus separate nested foreign stacks, and
@@ -102,17 +102,26 @@ cross-Runtime compact-handle sharing remains impossible. `CaptureDomain` separat
 collector domain and persists every member, internal function/global/table edge,
 memory32/memory64 and tag aliases, typed live passive roots, and one stable-ID heap
 graph in `WGDN` v3 with strict v1/v2 loading; restore publishes the complete member
-slice only after transactional graph reconstruction. Codec v34 persists helper admission and
-the 16-byte `v128` storage contract, but never compact handles.
+slice only after transactional graph reconstruction. Codec v35 persists helper admission,
+the required native-GC ABI version, and the 16-byte `v128` storage contract, but never compact handles.
 Snapshot v4 persists reachable local-global object graphs with stable IDs and
 two-pass cycle/sharing reconstruction; snapshot v5 adds one owned local
 collector-reference table, while snapshot v6 adds multiple heterogeneous local
 tables with indexed lengths, barriers, sharing, and exact structural root
 validation. AMD64 final scalar struct/array accesses and initialized final-struct
-allocation use collector native ABI v3: a stable allocation/collection/card-refreshed
-view at basedata offset 280 validates handle, space, object extent, canonical type,
-array bounds, remembered membership, any existing object-card interval, and the
-transactional native-allocation epoch before direct heap access. Final abstract
+allocation use collector native ABI v6. Artifact loading validates the Go/native
+layout and v35 records the required ABI; instantiation validates the immutable
+instance view, local canonical-type map, collector identity, collector version, and
+handle stride before publishing basedata offset 280. Native accesses then trust those
+immutable facts while reloading and validating mutable handle ranges/liveness, heap
+space and backing extents, object extents and canonical types, array bounds,
+remembered membership, any existing object-card interval, and the transactional
+native-allocation epoch. A module-owned 229-byte noncollecting resolver leaf replaces
+repeated inline resolver bodies at modules with at least two candidate sites; one-site
+modules retain inline code. Bounded straight-line reuse may retain one derived object
+address only while its unchanged compact local remains the root and no call,
+allocation, collection, host transition, control/EH edge, local mutation, or other
+invalidating opcode can intervene. Final abstract
 reference stores may bypass helpers only when no metadata growth is required;
 vector, non-final, bulk, cardless/unremembered, and Tiny barrier paths remain
 helper-bound. Arm64 builds lower struct/array/i31 and dynamic cast/test helpers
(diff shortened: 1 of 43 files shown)


jtenner:
CI note: the initial Darwin/arm64 job timed out on August 10, 2026 in the pre-existing TestThreadsDistinctInstancesOverlapInNativeExecution after 11 minutes. Both Darwin/arm64 Core 3 conformance jobs and all four GC-hardening jobs passed. The failed job has been re-run.

jtenner:
CI follow-up: the re-run Darwin/arm64 job passed in 3m26s, including build/test, guard-page traps, corpus matrix, and SIMD execution. All required CI jobs are now green.

jtenner:
Tertiary audit follow-up pushed in `adf40927`, `2c0ed7e9`, and `d7e02d7a`.

Findings fixed:
- recursive-group GC reference parameters now seed flattened type indexes rather than group-local indexes;
- `wagodebug` immutable native-view validation now runs after the entering instance rebinds `JobMemory`, including the private prepared path;
- resolver crossover scanning counts actual typed direct-scalar candidates rather than helper-bound/reference sites;
- single-function modules with reuse enabled lower inline first, then select the shared island only when at least two real resolutions remain (repeated-object fixture: 452 bytes; eight distinct objects still share at 949 vs 1,806 bytes inline);
- reused `ModuleStats` no longer retain stale module attribution;
- AMD64 and ARM64 AST/byte hint scans now classify GC/atomic helper calls and EH state consistently;
- module relocation patching now validates entry/site/target tables and uses checked signed `rel32` arithmetic.

Qualification refreshed:
- AMD64 backend and `./src/core/...` pass;
- focused release and `wagodebug` Wago tests pass;
- focused race tests pass;
- Linux/Darwin ARM64 test binaries cross-compile;
- runtime medians: 317.0 ns/op default, 341.85 ns/op no reuse, 340.6 ns/op inline/no reuse, all 0 B/op and 0 allocs/op;
- stripped TinyGo `wago_runtime`: 2,106,840 bytes vs 2,096,928 baseline (+9,912, +0.473%).

CI has restarted for head `d7e02d7a`.

jtenner:
CI follow-up: `71aafdb6` preserves the exact helper-free `ref.test`/`ref.cast` leaf classification while retaining call-safe hints for generic helper admission. This restores the pinned type-subtyping product sizes that failed Linux/race CI. Focused normal and race tests pass; refreshed stripped TinyGo size is 2,106,824 bytes (+9,896, +0.472% vs baseline). CI has restarted.

jtenner:
Final tertiary-audit qualification is green at head `71aafdb6`. All required CI checks passed: Linux/Darwin/Windows, amd64/arm64, Core v2/v3 explicit and signal modes, race, TinyGo, coverage, build size, lint, corpus provenance, and all four GC-hardening jobs. PR remains mergeable. The only local worktree entries are the pre-existing untracked `install` and `testdata/`, left untouched.

jtenner:
@JairusSW Could you please review this when you have a chance? The issue #307 implementation and tertiary audit are complete, all required CI checks are green, and the latest interleaved A/B benchmarks show a 3.11% runtime geomean improvement, an 11.47% improvement on eight repeated struct gets, 15.96% lower compile time, and 74.85% less generated code for that fixture.

PR #376, opened 2026-08-10
component: add component-model runtime

JairusSW:
## What changed

- adds a Component Model decoder, canonical ABI implementation, typed host linking, composition, resources, and synchronous/async task substrate
- adds structured core-import metadata and an explicit synchronous-host-call instantiation option for adapter modules
- makes nested host re-entry safe by isolating control frames, native stacks, trap cells, and call scratch
- supports memoryless embedded core modules and owned host function references used through component adapter tables

## Why

WASI Preview 2 guests are WebAssembly Components, not flat core modules. Real preview2 adapters also re-enter parked instances through indirect host calls; sharing the original native control frame corrupted the suspended activation and could crash the process.

## Impact

This introduces the public `src/component` API and the runtime substrate needed by `wago-org/wasi/p2`. Existing core-Wasm APIs remain compatible. Nested host callbacks now have explicit `HostTrap` handling and safe same-instance re-entry.

## Validation

- `go test ./src/component/... -count=1`
- targeted host re-entry, GC-root, indirect-funcref, and cross-instance context tests with `-count=3`
- real `wasi:cli/command` component execution in the companion WASI repository

The full `src/wago` suite was also audited: environment-dependent failures remain limited to absent external spec fixtures, the installed `wat2wasm` lacking table64, and one threads timing test that passes repeatedly in isolation.


JairusSW:
Superseded by #377 with the same implementation plus generated-facade, TinyGo-boundary, and feature-documentation fixes. GitHub Actions did not schedule synchronize/reopened runs for this draft head, so #377 was opened to obtain a clean CI run on the corrected SHA.

PR #378, opened 2026-08-10
component: add component-model runtime

JairusSW:
## What changed

- standardizes the narrow, revocable execution-model capability as `core.runtime` and exposes it through `PluginCoreRuntime` / `Registry.CoreRuntime()`: `Compile`, `Instantiate`, and `NewHostFuncRef`
- adds versioned typed plugin services with transactional dependency ordering, duplicate/missing-provider rejection, and shutdown revocation
- adds structured core-import metadata and explicit synchronous-host-call support for adapter modules
- makes nested host re-entry safe by isolating control frames, native stacks, trap cells, and call scratch
- supports memoryless embedded core modules and owned host function references used through component adapter tables
- keeps the Component Model implementation out of Wago core; it now lives in the standalone [wago-org/component-model](https://github.com/wago-org/component-model) plugin

## Architecture and DCE

Core Wago remains a core-Wasm runtime. This PR supplies only the capability and service substrate required by execution-model plugins. The standalone Component Model plugin receives `core.runtime` and publishes `wago-org/component-model/runtime/v1`; world plugins such as WASI P2 consume that service and receive no core authority.

The Component Model package is not imported by Wago root or its default CLI, so builds that do not select the plugin do not retain it.

## Security model

- capability access is inactive before transactional commit and revoked before runtime shutdown completes
- the privileged handle exposes a deliberately narrow engine interface, not `*Runtime`
- missing services, duplicate providers, type mismatches, dependency cycles, and ungranted capabilities fail closed before runtime mutation
- service references become inactive during reverse-order dependency teardown
- nested host callbacks use isolated native execution state and explicit trap propagation

## Validation

- Linux, Darwin, and Windows on amd64/arm64
- race detector, TinyGo, lint/staticcheck, coverage, build size, and Core v2/v3 conformance
- repeated same-instance concurrency and nested/cross-instance host re-entry tests
- exact `resource.action` capability naming with no parent, wildcard, or inheritance semantics
- capability denial, transactional service resolution, activation, runtime ownership, and shutdown revocation tests
- default Wago dependency graph excludes the standalone component plugin

Standalone implementation: https://github.com/wago-org/component-model
Companion WASI P2 integration: https://github.com/wago-org/wasi/pull/10


src/wago/api.go
```diff
@@ -3990,6 +3990,13 @@ func (in *Instance) InvokeContext(ctx context.Context, export string, args ...ui
 }
 
 func (in *Instance) invoke(export string, args []uint64, cancel context.Context) ([]uint64, error) {
+	if isNativeActive(in) {
```
JairusSW: **[P1] Do not identify nested re-entry with an instance-global active bit.** `isNativeActive(in)` is true while *any* goroutine is inside a host callback for this instance. During that callback the native lease has deliberately been released, so an unrelated goroutine may call `Invoke` on the same instance and enter this branch. It is then treated as a nested call and `prepareHostReentryState` swaps shared `in.eng`, control frames, serialized args/results, trap state, and native-context state. If both invocations reach host callbacks, the first can resume against state installed by the second, producing races or corrupt result/trap propagation. `beginInvocation` only tracks lifetime; it does not serialize these mutations.

Re-entry needs a call-chain-scoped token passed through `InvokeFromHost` (or equivalent), not a process-global map keyed only by `*Instance`. Unrelated public invocations should either be serialized at the instance boundary or use fully independent execution state. A regression can park invocation A in a host callback, start unrelated invocation B on another goroutine, let B enter its callback, then release A first; under `-race`, neither call should touch the other's control/trap/result state.

JairusSW (commented):
I found one correctness blocker in the new nested host re-entry mechanism. The service/capability planning and shutdown-order changes otherwise look internally consistent in the paths reviewed so far.

src/wago/api.go
JairusSW: Addressed in cf982f32. Public calls on an Instance are now serialized across parked host callbacks, while InvokeFromHost carries a callback-scoped invocation identity and only prepares isolated re-entry state when that same call chain owns the parked activation. The regression parks call A, starts unrelated call B, verifies B cannot enter until A completes, and rejects an expired callback capability. Focused re-entry/cross-instance tests and the regression pass under -race; Instance remains 832 bytes and steady-state invocation remains allocation-free.

PR #379, opened 2026-08-10
amd64: cost high-pressure associative tree covers

JairusSW:
## Summary

- expand AMD64 associative tree covering to every bounded Valent tree instead of stopping at eight leaves
- cache Sethi-Ullman pressure labels and cover need-four destination-hinted trees
- preserve repeated borrowed local/global destination reads once in a pinned scratch register while retaining the ordinary path for owned-register aliases
- add focused direct, nested, overflow, ownership, fallback, and kill-switch execution coverage
- document the full tree-optimization research and measured decisions

## Results

Combined versus main:

- 620 explicit-mode destination covers
- 4,529 fewer native bytes, 166 fewer spills, and 6 fewer reloads
- 552 guard-mode destination covers
- 652 fewer native bytes, 14 fewer spills, and 2 fewer reloads
- runnable corpus codegen statistics unchanged, so no execution-time win is claimed

The repeated-alias extension alone adds 305 explicit and 250 guard covers. Twelve interleaved pinned SQLite/Ruby samples leave ordinary backend/full compilation flat at 0.9994x geomean; ten forced-worker Ruby samples are flat at 1.0012x. Allocation volume is unchanged.

Rejected experiments:

- need-three destination policy: 0.35% affected-runtime geomean regression
- associative integer multiplication: zero new covers across all 64 corpus modules

## Validation

- full Linux/AMD64 backend suite passed on Ryzen 7 7800X3D
- AMD64 vet passed
- explicit corpus pipeline and differential execution passed
- guard-page corpus pipeline and differential execution passed
- prior full GitHub Actions run passed; fresh PR CI is running for the updated head

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/treecover.go
+++ b/src/core/compiler/backend/railshot/amd64/treecover.go
@@ -2,19 +2,40 @@
 
 package amd64
 
-const maxAssociativeLeaves = 8
+const minAssociativeDestNeed = 4
 
-// tryAssociativeTree covers a small, trap-free tree of one associative integer
-// operation as a single accumulator. The cover is fixed-size and allocation-free.
-// Destination-hinted trees keep the established local-sink alias handling.
+// tryAssociativeTree covers a bounded, trap-free tree of one associative integer
+// operation as a single accumulator. The existing Valent height bound is its
+// fuel: unlike the old fixed eight-leaf array, every tree the stack can defer is
+// eligible, and the two bounded walks allocate no scratch storage. Destination-
+// hinted trees keep the established local-sink alias handling.
 func (f *fn) tryAssociativeTree(node *elem, dest Reg) Reg {
-	if dest != regNone || !associativeOp(node.op) || treeRegisterNeed(node) < 3 || !treeAccumulatorSafe(node) {
+	requestedDest := dest != regNone
+	need := treeRegisterNeed(node)
+	if !associativeOp(node.op) || need < 3 {
 		return regNone
 	}
-	var leaves [maxAssociativeLeaves]*elem
-	n := 0
-	if !collectAssociativeLeaves(node, node.op, node.typ, leaves[:], &n) || n < 3 {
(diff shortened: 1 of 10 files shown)


JairusSW:
The merge was rolled back by restoring `main` to its pre-merge commit. Continued review is in #380.

src/core/compiler/backend/railshot/amd64/treecover.go
```diff
@@ -94,15 +201,38 @@ func associativeOp(op wOp) bool {
...
+		f.applyAssociativeLeaves(e.arg0, op, typ, first, dest)
+		f.applyAssociativeLeaves(e.arg1, op, typ, first, dest)
+		return
+	}
+	if e == first {
+		return
+	}
+	if e.isDeferred() {
+		f.condense(e, regNone)
+		f.pinned = f.pinned.add(dest)
```
jtenner, line 235: **[P1] Preserve the repeated-alias scratch pin across nested condensation.** `f.condense(e, regNone)` can clear a pin that was established by the outer cover. For example, after repeated destination reads are rewritten to `aliasCopy`, an integer-compare leaf can use that borrowed register in `condenseCompare`; that path does `pinned.add(L)` and later `pinned.remove(L)`. Because `pinned` is a bitmask rather than a reference count, the remove also drops this outer `aliasCopy` pin. This helper restores only `dest`, so a later deferred leaf may allocate and overwrite `aliasCopy` while still-later rewritten leaves read it, producing incorrect code. Please carry `aliasCopy` (or a protected-pin mask) into this walk and restore all outer pins after each nested `condense`, or make nested pin scopes preserve the prior mask. A regression should put repeated destination reads on both sides of a compare leaf followed by another allocating subtree, so the saved old value must remain live through multiple leaves.

jtenner (commented):
Found a blocking correctness issue in repeated-alias pin preservation. The inline comment includes the failure mechanism and a focused regression shape.

PR #380, opened 2026-08-10
amd64: cost high-pressure associative tree covers

JairusSW:
## Summary

- expand AMD64 associative tree covering to every bounded Valent tree instead of stopping at eight leaves
- cache Sethi-Ullman pressure labels and cover need-four destination-hinted trees
- preserve repeated borrowed local/global destination reads once in a pinned scratch register while retaining the ordinary path for owned-register aliases
- add focused direct, nested, overflow, ownership, fallback, and kill-switch execution coverage
- document the full tree-optimization research and measured decisions

## Results

Combined versus main:

- 620 explicit-mode destination covers
- 4,529 fewer native bytes, 166 fewer spills, and 6 fewer reloads
- 552 guard-mode destination covers
- 652 fewer native bytes, 14 fewer spills, and 2 fewer reloads
- runnable corpus codegen statistics unchanged, so no execution-time win is claimed

The repeated-alias extension alone adds 305 explicit and 250 guard covers. Twelve interleaved pinned SQLite/Ruby samples leave ordinary backend/full compilation flat at 0.9994x geomean; ten forced-worker Ruby samples are flat at 1.0012x. Allocation volume is unchanged.

Rejected experiments:

- need-three destination policy: 0.35% affected-runtime geomean regression
- associative integer multiplication: zero new covers across all 64 corpus modules

## Validation

- full Linux/AMD64 backend suite passed on Ryzen 7 7800X3D
- AMD64 vet passed
- explicit corpus pipeline and differential execution passed
- guard-page corpus pipeline and differential execution passed
- prior full GitHub Actions run passed; fresh PR CI is running for the updated head

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/treecover.go
+++ b/src/core/compiler/backend/railshot/amd64/treecover.go
@@ -2,19 +2,40 @@
 
 package amd64
 
-const maxAssociativeLeaves = 8
+const minAssociativeDestNeed = 4
 
-// tryAssociativeTree covers a small, trap-free tree of one associative integer
-// operation as a single accumulator. The cover is fixed-size and allocation-free.
-// Destination-hinted trees keep the established local-sink alias handling.
+// tryAssociativeTree covers a bounded, trap-free tree of one associative integer
+// operation as a single accumulator. The existing Valent height bound is its
+// fuel: unlike the old fixed eight-leaf array, every tree the stack can defer is
+// eligible, and the two bounded walks allocate no scratch storage. Destination-
+// hinted trees keep the established local-sink alias handling.
 func (f *fn) tryAssociativeTree(node *elem, dest Reg) Reg {
-	if dest != regNone || !associativeOp(node.op) || treeRegisterNeed(node) < 3 || !treeAccumulatorSafe(node) {
+	requestedDest := dest != regNone
+	need := treeRegisterNeed(node)
+	if !associativeOp(node.op) || need < 3 {
 		return regNone
 	}
-	var leaves [maxAssociativeLeaves]*elem
-	n := 0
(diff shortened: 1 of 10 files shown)


JairusSW:
Replacement for #379 after its premature merge was rolled back from `main`.

src/core/compiler/backend/railshot/amd64/treecover.go
```diff
@@ -94,15 +201,38 @@ func associativeOp(op wOp) bool {
...
+		f.applyAssociativeLeaves(e.arg0, op, typ, first, dest)
+		f.applyAssociativeLeaves(e.arg1, op, typ, first, dest)
+		return
+	}
+	if e == first {
+		return
+	}
+	if e.isDeferred() {
+		f.condense(e, regNone)
+		f.pinned = f.pinned.add(dest)
```
jtenner: **[P1] Restore every outer pin after condensing a rewritten leaf.** This is the live-PR carryover of the finding on #379. `aliasCopy` is pinned before the walk, but a nested leaf that reads it can temporarily add and then remove the same bit. For example, `condenseCompare` uses `pinned.add(L)` / `pinned.remove(L)` around a borrowed left operand. Since `regMask` is not reference-counted, that removal clears this cover's outer `aliasCopy` pin. This code restores only `dest`; the next allocating leaf can therefore reuse and overwrite `aliasCopy`, and a later rewritten destination read observes the clobbered value. Snapshot the protected pin mask before each nested `condense` and union it back afterward (or pass and explicitly restore both `dest` and `aliasCopy`). Please add an execution regression with: an initial destination seed, a rewritten destination read inside a compare, another deferred leaf that allocates a GPR, and a later rewritten destination read.

jtenner (changes requested):
The replacement PR still contains the repeated-alias pin-lifetime miscompile identified on #379. This needs a protected-pin restoration and a regression before merge.

PR #381, opened 2026-08-10
amd64: sink comparisons into byte stores

JairusSW:
## Summary

- keep integer comparison results in flags when immediately consumed by `i32.store8`
- emit `setcc` into a scratch low byte without the dead `movzx`
- add an AMD64 optimization knob, explain counter, encoder coverage, execution tests, and a kill switch

## Measured result

Pinned Ryzen 7 7800X3D, CPU 7, `GOMAXPROCS=1`:

- corpus coverage: 173 explicit-mode and 165 guard-mode sites across 11 modules
- native code: -684 bytes explicit, -569 bytes guard
- `Exec/sieve.count`: 81.15 us/op to 81.01 us/op, 0.9983x, p=0.000, n=12
- `Compile/sqlite3` + `Compile/ruby`: 1.0001x geomean, no significant row; allocation volume effectively unchanged

The initial implementation regressed Ruby compilation by 1.83% because it refactored every comparison through an extra helper. This version restores the ordinary compare path exactly and pays only at matched store sinks.

## Validation

- `go test ./src/core/encoder/amd64 ./src/core/compiler/backend/railshot/amd64`
- `go vet ./src/core/compiler/backend/railshot/amd64`
- `go test ./src/core/compiler/optimization ./src/wago`
- `WAGO_CORPUS_TIMEOUT=20s make test-corpus` on AMD64: explicit and guard-page suites pass
- `WAGO_NO_STORE8_FLAGS=1` differential oracle exercised

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/memory.go
+++ b/src/core/compiler/backend/railshot/amd64/memory.go
@@ -615,6 +615,25 @@ func (f *fn) memStore(r *wasm.Reader, size int) error {
 		}
 		return nil
 	}
+	// An integer comparison immediately consumed by an 8-bit store needs only its
+	// low byte. Keep SETcc's upper-register garbage dead and omit MOVZX; the byte
+	// store cannot observe it. Pending loads were materialized above, preserving
+	// pre-store reads and trap order before this dedicated sink condenses the tree.
+	if top := f.s.back(); size == 1 && store8FlagsEnabled && isFusableCompare(top) && !top.typ.isFloat() {
+		vreg := f.allocReg(0)
+		f.pinned = f.pinned.add(vreg)
+		cc := f.condenseToFlags(top)
+		f.stats.reclassifyPeep("cmp-branch-fuse", "store8-flags")
+		f.a.SetccReg8(cc, vreg)
+		ea, eaOwned, _, disp := f.memAddr(off32, size, true)
+		f.a.StoreIdx(RBX, ea, vreg, disp, size)
+		f.pinned = f.pinned.remove(vreg)
+		if eaOwned {
+			f.release(ea)
+		}
+		f.release(vreg)
+		return nil
+	}
 	// Both the value and the address are immediate read-only uses here, so a
 	// pinned local feeds the store in place — no copy (nothing between the reads
 	// and the StoreIdx can write a local).
--- a/src/core/compiler/backend/railshot/amd64/memory.go
+++ b/src/core/compiler/backend/railshot/amd64/memory.go
@@ -620,7 +620,11 @@ func (f *fn) memStore(r *wasm.Reader, size int) error {
 	// store cannot observe it. Pending loads were materialized above, preserving
 	// pre-store reads and trap order before this dedicated sink condenses the tree.
 	if top := f.s.back(); size == 1 && store8FlagsEnabled && isFusableCompare(top) && !top.typ.isFloat() {
-		vreg := f.allocReg(0)
+		// condenseToFlags may recursively lower div/rem or a variable shift. Those
+		// paths temporarily claim and then unpin x86's fixed-role registers; because
+		// the pin mask is not reference-counted, nesting would drop this outer
+		// reservation. Keep the byte result in a neutral scratch instead.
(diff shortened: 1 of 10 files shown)


JairusSW (commented):
Found one blocking correctness issue under register pressure. The direct byte-store encoding, evaluation/trap ordering, stats reclassification, public knob, kill switch, and CI coverage otherwise look consistent.

src/core/compiler/backend/railshot/amd64/memory.go
```diff
@@ -615,6 +615,25 @@ func (f *fn) memStore(r *wasm.Reader, size int) error {
 		}
 		return nil
 	}
+	// An integer comparison immediately consumed by an 8-bit store needs only its
+	// low byte. Keep SETcc's upper-register garbage dead and omit MOVZX; the byte
+	// store cannot observe it. Pending loads were materialized above, preserving
+	// pre-store reads and trap order before this dedicated sink condenses the tree.
+	if top := f.s.back(); size == 1 && store8FlagsEnabled && isFusableCompare(top) && !top.typ.isFloat() {
+		vreg := f.allocReg(0)
```
JairusSW: **Blocking: keep this scratch out of fixed-role registers.**

`allocReg(0)` can return `RAX`, `RDX`, or `RCX` once the non-fixed pool is occupied. `condenseToFlags` may recursively lower a div/rem or variable shift; those paths claim the fixed register and later remove it from `f.pinned` (the pin set is a bitmask, not nested/ref-counted). That silently drops this outer reservation. After `SetccReg8`, `memAddr` can then allocate the same register for the address or bounds scratch and overwrite the byte before `StoreIdx`.

A concrete shape is: keep enough pending loads live to force `vreg = RAX`, then store `(x / y) < z` through a non-pinned address. `condenseDivRem` clears the RAX pin, and `memAddr` can reuse RAX and clobber AL.

Please avoid the fixed-role set here (for example, `allocReg(maskOf(RAX, RDX, RCX))`) or otherwise re-establish a tracked reservation after condensing, and add a register-pressure regression with a nested fixed-register operand.

src/core/encoder/amd64/asm_base_test.go
```diff
@@ -107,6 +107,7 @@ func TestIntegerMemoryAndControlEncodings(t *testing.T) {
 		{"multiply high", func(a *Asm) { a.IMulHigh(R8, false) }, []byte{0x41, 0xf7, 0xe8}},
 		{"conditional move", func(a *Asm) { a.Cmovcc(CondGE, R8, R9, true) }, []byte{0x4d, 0x0f, 0x4d, 0xc1}},
 		{"setcc register", func(a *Asm) { a.SetccReg(CondNE, R8) }, []byte{0x41, 0x0f, 0x95, 0xc0, 0x45, 0x0f, 0xb6, 0xc0}},
+		{"setcc low byte", func(a *Asm) { a.SetccReg8(CondNE, R8) }, []byte{0x41, 0x0f, 0x95, 0xc0}},
```
JairusSW, line 110: **Test the distinct bare-REX path as well.** R8 covers `REX.B`, but `SetccReg8` has a separate mandatory `0x40` prefix path for RDI/RSI/RBP, which are high-priority allocatable registers. Without that prefix the ModRM byte names BH/DH/CH instead of DIL/SIL/BPL. An `RSI` case (`40 0f 95 c6`) would pin down the live 4–7 register path.

src/core/compiler/backend/railshot/amd64/memory.go
JairusSW: Fixed in 7bac77de. The store8 flag scratch now excludes RAX, RDX, and RCX, so nested div/rem and variable-shift fixed-register pinning cannot make the deferred address reuse the flag byte. Added a deterministic pressure regression that produced result=137 before the fix and now stores 1.

src/core/encoder/amd64/asm_base_test.go
JairusSW: Fixed in 7bac77de. Added explicit coverage for the low-byte mandatory bare REX path using RSI/SIL, asserting 40 0f 95 c6.

PR #385, opened 2026-08-11
fix memory bounds after host reentry

JairusSW:
Fix stale linear-memory bounds after synchronous host re-entry grows shared memory.

- reload the cached memory size after sync host calls
- reload it after restoring dynamic cross-instance caller context
- cover nested guest re-entry + memory.grow with an amd64 runtime regression
- keep the arm64 path in parity

Validation:
- focused regression passes 20 consecutive runs
- amd64 compiler package passes
- arm64 compiler/runtime packages cross-compile
- the WASI Preview 2 corpus passes all 35 Wasmtime-reference fixtures with this fix

The broader src/wago package includes staged spec tests that require the absent tests/spec-v3 checkout; those fail at corpus discovery in this worktree.

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/call.go
+++ b/src/core/compiler/backend/railshot/amd64/call.go
@@ -14,6 +14,16 @@ import (
 	"github.com/wago-org/wago/src/core/runtime/abi"
 )
 
+// refreshCachedMemoryBoundAfterExternalCall reestablishes the memory-zero
+// bounds cache after a continuation resumes from code outside this compilation
+// context. The external code may have grown an aliased memory, and a
+// cross-instance continuation must read from the restored caller basedata.
+func (f *fn) refreshCachedMemoryBoundAfterExternalCall() {
+	if f.memSizeReg != regNone {
+		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
+	}
+}
+
 // regABIEnabled turns on the register-based internal-call ABI (default on;
 // WAGO_Amd64_NOREGABI=1 forces the wrapper ABI everywhere, for A/B measurement).
 var regABIEnabled = os.Getenv("WAGO_Amd64_NOREGABI") != "1"
@@ -629,9 +639,7 @@ func (f *fn) emitTailDynamicImportJump(ft *wasm.CompType, b ImportBinding) {
 	f.a.Load64(RBX, RSP, 0)
 	f.a.Load64(R10, RSP, 8)
 	f.copyInstanceContext(RBX, R10)
-	if f.memSizeReg != regNone {
-		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
-	}
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	f.deriveModuleGlobals()
 	if len(ft.Results) > 0 {
 		if isFloatValType(ft.Results[0]) {
@@ -707,9 +715,7 @@ func (f *fn) emitTailCrossDirectJump(ft *wasm.CompType, b ImportBinding) {
 	trampoline := f.a.Len()
 	f.a.PatchRel32(trampolineSite, trampoline)
 	f.a.Load64(RBX, RSP, 0)
-	if f.memSizeReg != regNone {
-		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
-	}
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	f.deriveModuleGlobals()
 	if len(ft.Results) > 0 {
 		if isFloatValType(ft.Results[0]) {
@@ -1142,6 +1148,10 @@ func (f *fn) callHostSync(importIdx int, ft *wasm.CompType) error {
 		}
 		value.st.gcRoot = gcFrameRefType(f.m, ft.Results[j])
 	}
+	// Arbitrary host code can synchronously re-enter this instance and grow its
+	// memory. Reload after reconstructing the operand stack so the continuation
+	// cannot retain the parked activation's pre-call size.
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	return nil
 }
 
@@ -1409,6 +1419,10 @@ func (f *fn) emitCrossInstanceCall(b ImportBinding, ft *wasm.CompType) error {
 	if b.Dynamic {
 		f.copyInstanceContext(RBX, R9)
 	}
+	// A dynamic target may be arbitrary host code that synchronously re-enters
+	// this caller and grows its memory. R15 was saved before that call, so reload
+	// from the restored caller context instead of reviving the stale bound.
+	f.refreshCachedMemoryBoundAfterExternalCall()
 
 	f.reloadLocalsForCall() // non-STACK_REG model only
 	f.deriveModuleGlobals() // cross-instance callee may have written shared global cells
@@ -2370,9 +2384,7 @@ func (f *fn) emitTailCrossWrapperJump(ft *wasm.CompType) {
 	f.a.Load64(RBX, RSP, 0)
 	f.a.Load64(R10, RSP, 8)
 	f.copyInstanceContext(RBX, R10)
-	if f.memSizeReg != regNone {
-		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
-	}
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	f.deriveModuleGlobals()
 	if len(ft.Results) > 0 {
 		if mtOf(ft.Results[0]).isFloat() {
--- a/src/core/compiler/backend/railshot/arm64/call.go
+++ b/src/core/compiler/backend/railshot/arm64/call.go
@@ -15,6 +15,16 @@ import (
 	"github.com/wago-org/wago/src/core/runtime/abi"
 )
 
+// refreshCachedMemoryBoundAfterExternalCall reestablishes the memory-zero
+// bounds cache after a continuation resumes from code outside this compilation
+// context. The external code may have grown an aliased memory, and a
+// cross-instance continuation must read from the restored caller basedata.
+func (f *fn) refreshCachedMemoryBoundAfterExternalCall() {
+	if f.memSizeReg != regNone {
+		f.ld64(f.memSizeReg, linMemReg, -int32(bdCurBytes))
+	}
+}
+
 // regABIEnabled turns on the register-based internal-call ABI (default on;
 // WAGO_ARM64_NOREGABI=1 forces the wrapper ABI everywhere, for A/B measurement).
 var regABIEnabled = os.Getenv("WAGO_ARM64_NOREGABI") != "1"
@@ -432,9 +442,7 @@ func (f *fn) emitTailDynamicImportJump(ft *wasm.CompType, b ImportBinding) error
 	f.ld64(X11, SP, 16)
 	f.copyInstanceContext(X10, X11)
 	f.a.MovReg64(linMemReg, X10)
-	if f.memSizeReg != regNone {
-		f.ld64(f.memSizeReg, linMemReg, -bdCurBytes)
-	}
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	f.deriveModuleGlobals()
 	f.derivePinnedGlobals()
 	if len(ft.Results) > 0 {
@@ -686,9 +694,7 @@ func (f *fn) emitTailDescriptorWrapperJump(ft *wasm.CompType) {
 	f.ld64(X11, SP, 16)
 	f.copyInstanceContext(X10, X11)
 	f.a.MovReg64(linMemReg, X10)
-	if f.memSizeReg != regNone {
-		f.ld64(f.memSizeReg, linMemReg, -bdCurBytes)
-	}
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	f.deriveModuleGlobals()
 	f.derivePinnedGlobals()
 	if len(ft.Results) > 0 {
@@ -1064,6 +1070,10 @@ func (f *fn) callHostSync(importIdx int, ft *wasm.CompType) error {
 		}
 		value.st.gcRoot = f.tracksGCFrameRoots() && arm64GCFrameRefType(f.m, ft.Results[j])
 	}
+	// Arbitrary host code can synchronously re-enter this instance and grow its
+	// memory. Reload after reconstructing the operand stack so the continuation
+	// cannot retain the parked activation's pre-call size.
+	f.refreshCachedMemoryBoundAfterExternalCall()
 	return nil
 }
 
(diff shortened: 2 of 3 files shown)


src/wago/shared_memory_basedata_test.go
```diff
@@ -145,6 +146,55 @@ func TestSharedMemoryIndirectCallSwitchesPrivateContext(t *testing.T) {
...
+		(import "env" "reenter" (func $reenter (result i32)))
+		(memory (export "memory") 1 2)
+		(func (export "grow") (result i32)
+			(memory.grow (i32.const 1)))
+		(func (export "outer") (result i32)
+			(call $reenter)
+			(i32.const 65536)
+			i32.mul
+			i32.const 42
+			i32.store
```
jtenner: **Blocker after rebasing onto current main:** this nested `instance.Invoke("grow")` no longer follows the finalized invocation contract. Calls on one `Instance` are serialized while the outer activation is parked in a host callback, so a plain nested `Invoke` can deadlock on the same instance gate. Please use `instance.InvokeFromHost(ctx, caller, "grow")` here and give the test a bounded context so a future re-entry regression fails instead of hanging the suite. I would also add a direct host-side `Memory.Grow` case because the invalidation boundary is arbitrary host execution, not only guest re-entry.

src/core/compiler/backend/railshot/amd64/call.go
```diff
@@ -1409,6 +1415,13 @@ func (f *fn) emitCrossInstanceCall(b ImportBinding, ft *wasm.CompType) error {
 	if b.Dynamic {
 		f.copyInstanceContext(RBX, R9)
 	}
+	// The imported target can synchronously enter Go and re-enter a module that
+	// grows this caller's shared memory. R15 was saved before that call, so reload
+	// the current size from the restored caller context instead of reviving the
+	// stale pre-call bound.
+	if f.memSizeReg != regNone {
+		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
```
jtenner: This reload looks necessary, but the PR has no regression that actually exercises this `emitCrossInstanceCall` continuation. The existing added test only covers a synchronous host import. Please add an A -> imported `InstanceExport` B case where A and B alias the same memory, B grows it, then A immediately touches the new page. Also add a non-aliased control case to prove the value is reloaded from the restored caller context. The comment should describe aliased-memory mutation generally; the callee can change the caller-visible bound directly, not only by re-entering through Go.

src/wago/host_execution.go
```diff
@@ -109,14 +110,86 @@ func (root *Instance) dispatchSynchronousHostCall(ctrl uintptr, importIdx uint32
...
+			// whose custom-context word names its original control frame. A nested
+			// host re-entry has a distinct live frame; republish that exact frame
+			// before resuming its parked native activation.
+			active.jm.SetCustomCtx(ctrl)
 		}
 	}()
+	// Only arbitrary host code can synchronously re-enter this instance. Marking
+	// the whole native activation made an unrelated concurrent Invoke look like
+	// host re-entry and swap the live engine and scratch buffers underneath it.
+	markNativeActive(active)
```
jtenner, line 131: I think the re-entry marker needs an activation identity, not only an `*Instance`. A native `InstanceExport` call can enter B without going through B's Go-level `invoke`, so B's host callback does not inherently have B's own public invocation token. Conversely, if B has an unrelated parked public invocation, an instance-scoped ID can be accidentally reused by a different native call chain. Please propagate the root invocation identity through native cross-instance activation/control-frame state and key `markNativeActive`/`InvokeFromHost` authorization from that exact activation. A regression should be genuinely native: `A.Invoke -> native import into B -> B HostFunc -> A.InvokeFromHost(...)`.

src/wago/host_execution.go
```diff
@@ -109,14 +110,86 @@ func (root *Instance) dispatchSynchronousHostCall(ctrl uintptr, importIdx uint32
...
+	if err := coreruntime.InitHostCtrlFrame(ctrl); err != nil {
+		_ = coreruntime.ReleaseEngine(eng)
+		return nil, err
+	}
+
+	outerEngine, outerCtrl := in.eng, in.ctrl
+	outerArgs, outerResults, outerTrap := in.serArgs, in.results, in.trap
+	outerResultVals := in.resultVals
+	outerInvokeCache, outerInvokeCacheNext := in.ic, in.icNext
+	outerInstructionState := in.instructionState
```
jtenner, line 156: This isolation works by temporarily swapping mutable invocation state directly on the shared `Instance` (`eng`, `ctrl`, args/results/trap, cache, instruction state). That creates a fragile lifetime boundary: a host callback can start `InvokeFromHost` on another goroutine, pass validation, return, and let the outer activation resume while the nested operation is still using/swapping these fields. It also means `Close`/cancellation can observe only whichever trap is currently installed. Please either add a callback-scope re-entry lease that prevents the host scope/outer continuation from ending until the nested call is fully restored, or move nested scratch into an explicit activation object. I would add adversarial tests for callback escape, Close during nested execution, and root cancellation while nested guest code is running.

jtenner (changes requested):
The cached-bound reload itself looks correct, but I don't think this branch is mergeable as-is. The PR is now 5 commits behind current main, and much of the 16-commit component/plugin stack in this branch has already landed separately on main in its finalized form. I strongly recommend rebuilding this PR from current main and porting only the unique bounds fix + focused regressions rather than conflict-resolving the old component stack in place.

Blocking items below:
- update the regression to use the finalized callback-scoped re-entry API;
- add a direct regression for the cross-instance reload path;
- make invocation identity activation-local/propagated for native cross-instance host re-entry;
- harden nested activation lifetime instead of relying on mutable Instance field swapping.

The first two are directly required for the memory-bounds patch. The latter two are runtime/re-entry correctness concerns exposed by this branch and should either be fixed here or explicitly split into prerequisite/follow-up PRs with the memory fix scoped so it does not claim unsupported native cross-instance re-entry behavior.

src/core/compiler/backend/railshot/amd64/call.go
```diff
@@ -1142,6 +1142,12 @@ func (f *fn) callHostSync(importIdx int, ft *wasm.CompType) error {
 		}
 		value.st.gcRoot = gcFrameRefType(f.m, ft.Results[j])
 	}
+	// Arbitrary host code can synchronously re-enter this instance and grow its
+	// memory. Reload after reconstructing the operand stack so the continuation
+	// cannot retain the parked activation's pre-call size.
+	if f.memSizeReg != regNone {
+		f.a.Load64(f.memSizeReg, RBX, -bdCurBytes)
```
jtenner: The load itself looks right, but I would centralize this as something like `refreshCachedMemoryBoundAfterExternalCall()` (or an invalidation helper) and use it from every external continuation that can resume the current compilation context. Tail cross-instance trampolines already reload the bound; the ordinary host/cross-instance paths drifted. A shared helper makes that invariant auditable and reduces the chance of another path missing the refresh.

src/core/compiler/backend/railshot/arm64/call.go
```diff
@@ -1064,6 +1064,12 @@ func (f *fn) callHostSync(importIdx int, ft *wasm.CompType) error {
 		}
 		value.st.gcRoot = f.tracksGCFrameRoots() && arm64GCFrameRefType(f.m, ft.Results[j])
 	}
+	// Arbitrary host code can synchronously re-enter this instance and grow its
+	// memory. Reload after reconstructing the operand stack so the continuation
+	// cannot retain the parked activation's pre-call size.
+	if f.memSizeReg != regNone {
+		f.ld64(f.memSizeReg, linMemReg, -int32(bdCurBytes))
```
jtenner: Source parity with AMD64 is good, but cross-compilation is not enough evidence for this runtime bug. Please add ARM64 execution or code-generation coverage that proves the cached bound reload occurs *after* the caller context is restored. A native ARM64 test runner is ideal; otherwise a backend/disassembly assertion around this continuation would still guard against a future ordering regression.

jtenner (commented):
Backend/performance follow-up for the minimal bounds fix. I think the conservative reload is the right first correction, but this bug existed because the same external-continuation invariant is implemented independently in several paths. Before calling this complete, I would tighten the invariant and add evidence around its cost.

Suggested coverage matrix:
- `memory.size` immediately after return;
- scalar load/store and `memory.copy`/`memory.fill` on the grown page;
- failed grow at max;
- repeated host-call growth;
- aliased vs non-aliased cross-instance memory;
- explicit-bounds mode selected in the regression so it definitely exercises `memSizeReg`;
- multi-memory/memory64 where those paths use a cached memory-zero bound.

Suggested benchmarks (with `ReportAllocs`): synchronous no-op host call, cross-instance call, and host-call-followed-by-memory-access. Also report generated-code byte delta for a module with many host call sites. If the unconditional reload is measurable when no memory operation follows, a later lazy invalidation/reload scheme could recover that cost.

jtenner (commented):
Additional hardening notes from the broader 16-commit branch. These should **not** block a clean minimal bounds-fix PR if this stale component/plugin history is dropped during rebase; they are follow-up issues for the runtime/component surface itself.

1. **Nested cancellation should inherit the root invocation.** `InvokeFromHost(context.Background(), ...)` can currently detach a nested activation from the outer `InvokeContext` deadline because re-entry receives a fresh trap cell/engine. Treat the supplied context as an additional child constraint: effective cancellation should be parent cancellation OR the provided context.

2. **Close needs to interrupt all live activations, not whichever trap is currently installed.** Nested re-entry temporarily swaps `in.trap`; an outer parked trap is therefore not represented by the single field. Track live activation traps explicitly and arm them all on Close/root cancellation.

3. **Nested re-entry allocates heavily.** Each depth acquires an engine and allocates control/arg/result/trap/result-value buffers. Add `BenchmarkInvokeFromHostReentryDepth{1,2,4}` with `ReportAllocs`; if this path matters for component adapters, use a bounded activation pool/free list with retention limits rather than an unbounded `sync.Pool`.

4. **Global `nativeActive` tracking may become a contention point.** Every arbitrary host call touches a process-wide mutex/map while host callbacks can run concurrently. Once activation identity is explicit, consider moving active-depth/authorization to activation-local or per-instance state and benchmark the host-call path.

5. **Threads currently avoid a latent lock mismatch only by admission rules.** Threaded entries use `memoryDir.nativeMu`, while synchronous host dispatch releases/reacquires the process-wide `nativeExecutionMu`. Function imports are currently rejected for threaded modules, so this is unreachable today. Encode that invariant in a regression and replace the hardcoded global unlock with an exact execution-lease abstraction before threads gain host imports.

6. **Core runtime access revocation needs an operation lease.** Copying `rt` out from under an `RWMutex` and then invoking it permits a close/revoke race between validation and operation completion. Hold a read lease for the full Compile/Instantiate/NewHostFuncRef operation or track in-flight privileged operations; also make closed-runtime behavior consistent across all public entry points.

7. **Service registry hardening:** reject duplicate service names within one scratch registry, reject typed-nil providers (pointer/map/slice/interface/func/chan), and ensure failed binding cannot leave partial runtime mutations. Add interface-assignability and rollback tests.

8. **Cleanup errors:** `prepareHostReentryState` panics if releasing the nested engine fails. Cleanup failure should be joined with the primary error/trap rather than masking it.

These are mostly architecture follow-ups, but they are worth tracking because the component model exercises exactly these nested execution boundaries.

src/core/runtime/engine.go
```diff
@@ -274,7 +274,7 @@ func (e *Engine) callWithHostLoop(code uintptr, serArgs []byte, linMemBase uintp
 			for k := 0; k < nres; k++ {
 				resBuf[k] = 0
 			}
-			host(ctrlPtr, imp, argBuf[:n], resBuf)
+			host(ctrlPtr, imp, argBuf[:n], resBuf[:nres])
```
jtenner, line 277: Nice reduction from the full 64-slot result buffer, but please cap the slices as well as their lengths: `argBuf[:n:n]` and `resBuf[:nres:nres]`. With the current two-index form a hostile/buggy HostFunc can reslice up to the backing array capacity, inspect stale argument/result slots from prior calls, or write beyond its declared result arity. Full slice expressions harden this with no allocation cost; add a malicious-host regression that verifies `cap(args)==len(args)` and `cap(results)==len(results)`.

src/wago/runtime.go
```diff
@@ -268,6 +296,22 @@ func (rt *Runtime) Compile(wasmBytes []byte) (*Module, error) {
 		return nil, err
 	}
 	mod := rt.buildModule(c)
+	// The historical Imports key joins module and field with a dot. Both Wasm
+	// names may themselves contain dots, so recover the exact pair from the
+	// validated source for structured metadata and component-model linking.
+	// The flat key remains unchanged for backwards-compatible Imports lookup.
+	if decoded, decodeErr := wasm.DecodeModule(source); decodeErr == nil {
```
jtenner, line 303: This fixes display metadata for dotted function names by decoding the Wasm a second time, but it leaves the underlying flat import identity (`module + "." + name`) ambiguous and adds another O(module-size) decode/allocation to every `Runtime.Compile`. `("a.b","c")` and `("a","b.c")` still collide. I would preserve a structured `{Module, Name}` import key in `Compiled` during the *first* decode, serialize it in artifacts, and keep the flat string only as a compatibility adapter. Please also cover `Runtime.Module(precompiled)` and metadata parity; this post-pass recovery only exists on the source compile path.

JairusSW:
Rebuilt this PR from current `main` as one focused commit (`88cf1024`); the stale component/plugin stack is gone.

Implemented:
- centralized cached memory-bound refreshes for AMD64 and ARM64 external continuations;
- callback-scoped `InvokeFromHost` with a 5-second regression timeout;
- explicit-bounds coverage for `memory.size`, scalar load/store, `memory.fill`/`memory.copy`, repeated growth, and failed growth at max;
- a benchmark matrix for no-op host calls, host-then-memory access, cross-instance control calls, and 64 host call sites.

Scope resolutions:
- Wago has no public host-side `Memory.Grow` API, so this PR does not add or claim that separate API;
- native `A -> InstanceExport B -> HostFunc -> A.InvokeFromHost` remains out of scope until activation identity is propagated through native cross-instance activation;
- the stale `host_execution.go`, `engine.go`, and `runtime.go` review surfaces are no longer in this diff.

Local proof:
- `go test ./src/wago -count=1`
- `go test ./src/core/compiler/backend/railshot/arm64 -count=1`
- `GOARCH=amd64 go test ./src/core/compiler/backend/railshot/amd64 -count=1`
- `GOARCH=amd64 go test ./src/wago -run '^TestHostReentryRefreshesMemorySizeAfterNestedGrow$' -count=1`
- `go test -race ./src/wago -run '^TestHostReentryRefreshesMemorySizeAfterNestedGrow$' -count=1`

Darwin/ARM64 benchmark comparison against `77b89c17` (5 runs, median):
- host no-op: 224.4 ns -> 227.8 ns; allocations unchanged at 48 B / 1 alloc;
- host then load: 226.7 ns -> 232.8 ns; allocations unchanged at 48 B / 1 alloc;
- cross-instance control: 97.85 ns -> 100.8 ns; allocations unchanged at 0 B / 0 allocs;
- 64 host sites: 8282 ns -> 8539 ns; allocations unchanged at 3072 B / 64 allocs.

Generated code grows by exactly 8 bytes per external call site on ARM64: 464 B -> 472 B for one site and 17,096 B -> 17,608 B for 64 sites. The conservative reload is retained; lazy invalidation can be evaluated separately if this cost becomes material.

PR #419, opened 2026-08-11
railshot: schedule large straight-line regions without compiler IR

JairusSW:
## What changed

- scans validated Wasm bytes into a bounded Railshot-local value plan; production code no longer imports or builds `src/core/compiler/ir`
- keeps the general compiler IR strictly off the execution path, enforced by its package-boundary test
- schedules large call-free/control-free integer regions on AMD64 and ARM64 while retaining the direct compiler as the universal fallback
- admits integer add/subtract/multiply/logic, constant shifts/rotates, i64-to-i32 wrap, every integer load/store width, and ordered interleaved memory
- uses next-use eviction, dying-operand reuse, and architecture-specific emission
- keeps the newly merged compact/interval local optimizers on fallback functions and makes their frame/register contracts mutually exclusive with this scheduler

Differential switches:

- `WAGO_AMD64_NO_STRAIGHTLINE_SSA=1`
- `WAGO_ARM64_NO_STRAIGHTLINE_SSA=1`

The corpus scan admits `blake-as`, `blake-as-simd`, `blake3sum`, and eleven large Ruby functions. The latter two modules are compile-only in the current harness.

## Current-main execution

Guard-page mode, `GOMAXPROCS=1`, five 750 ms samples; 0 B/op and 0 allocs/op throughout. AMD64 is pinned to CPU 7.

| Host / corpus | direct | planned | latency delta |
|---|---:|---:|---:|
| Ryzen 7 7800X3D / blake-as | 689,978 ns/op | 523,334 ns/op | -24.2% |
| Ryzen 7 7800X3D / blake-as-simd | 538,766 ns/op | 454,792 ns/op | -15.6% |
| Apple M4 Max / blake-as | 402,594 ns/op | 379,398 ns/op | -5.8% |
| Apple M4 Max / blake-as-simd | 436,976 ns/op | 425,390 ns/op | -2.7% |

AMD64 geometric-mean latency improves 20.0%, equivalent to **1.250x throughput**. ARM64 geometric-mean latency improves 4.2% after #386 substantially improved the direct baseline.

Four unrelated executable corpora (`nbody`, `quicksort`, `sha256`, and both `json-as` paths) remain scheduler no-hits and serve as broad regression checks.

## Compile cost and footprint

Five 500 ms full-compile samples against the current direct fallback:

| Host / corpus | direct | planned | latency delta |
|---|---:|---:|---:|
| AMD64 / blake-as | 487,756 ns/op | 508,498 ns/op | +4.3% |
| AMD64 / blake-as-simd | 1,530,199 ns/op | 1,567,645 ns/op | +2.4% |
| ARM64 / blake-as | 331,942 ns/op | 338,114 ns/op | +1.9% |
| ARM64 / blake-as-simd | 1,106,301 ns/op | 1,116,121 ns/op | +0.9% |

Compared with the previous general-IR builder, the backend-local planner removes about **123 KB and 18 allocations per optimized compile**. Hot native function sizes remain 7,016/7,025 bytes on AMD64 and 4,228/4,268 bytes on ARM64.

## Validation

- `go test ./src/core/compiler/... ./src/core/runtime/...` on Darwin/ARM64 and Linux/AMD64
- `go vet ./src/core/compiler/... ./src/core/runtime/...`
- `go test -tags wago_guardpage ./src/core/compiler/backend/railshot/arm64 ./src/core/runtime/...`
- AMD64 admission/fallback and compact-frame exclusion tests run natively on `hub`
- explicit/guard differential execution and broad no-hit corpus A/Bs on both architectures


JairusSW:
Closing this rather than weakening Wago’s single-pass architecture. A bounded linear-cache experiment recovered only about 3% on scalar BLAKE and slightly regressed the SIMD corpus; the remaining gain depends on retaining whole-function value identities across local assignments. That is full SSA behavior, not an honest peephole. The branch is intentionally discarded.

Issue #319, opened 2026-08-04
[P1][TinyGC] Bound incremental work by slots/bytes and simplify the mutation barrier


jtenner:
## Context

Tiny `Step` currently processes one object, but one object can be a giant reference array, so work and pause time are not bounded. Root reset, color initialization, synchronous barrier drains, and sweeping can also perform unbounded work. Define incremental units in bytes/reference slots and make allocation debt drive pacing.

### Bounded marking

- [ ] Budget work in reference slots and/or payload bytes, not object count.
- [ ] Add resumable scan cursors for large structs/arrays.
- [ ] Split giant reference-array tracing across multiple `Step` calls.
- [ ] Incrementally enumerate persistent roots with cursors.
- [ ] Clear marks/colors logically with epoch or polarity rather than scanning every handle at cycle start.

### Bounded sweeping

- [ ] Sweep by page/span cursor with an explicit byte/block budget.
- [ ] Permit allocation from already swept regions while later regions remain unswept.
- [ ] Avoid synchronously draining the entire cycle merely because allocation occurs during sweep.
- [ ] Preserve newly allocated object coloring for mark, remark, and sweep states.

### Pacing

- [ ] Track allocation debt and perform collector work proportional to allocated bytes.
- [ ] Adapt the work multiplier when the collector falls behind or the heap nears exhaustion.
- [ ] Expose hard bounded knobs suitable for constrained/embedded products.

### Barrier design

- [ ] Evaluate a simpler proven incremental-update or SATB-style barrier against the current hybrid forward/backward barrier.
- [ ] Replace synchronous full graph drains from root/object barriers with a bounded gray queue where possible.
- [ ] Preserve root stores made during sweep and prevent reclamation of newly published objects.
- [ ] Keep no-barrier behavior while the collector is idle.

### Product variants

- [ ] Evaluate `TinyNonIncremental` for smallest binary/simplest collector and `TinyIncremental` for bounded pauses.
- [ ] Allow build-time exclusion of unused Tiny policy code.

## Benchmarks and gates

- Giant reference array, deep linked graph, mutation storm during each phase, allocation during sweep, root churn, and near-exhaustion pacing.
- Report p99/max step time, slots/bytes processed, barrier cost, cycle completion latency, throughput, metadata, and linked bytes.

## Acceptance criteria

- A configured Tiny step has an enforceable upper work bound independent of object size.
- Mutation barriers preserve the tri-color/root invariant under stress without unbounded synchronous tracing.
[...]

PR #420, opened 2026-08-11
gc: clear Tiny marks with compact epochs

jtenner:
## Summary

Continue #319 by replacing TinyGC's handle-wide color reset with one-byte epoch-encoded mark state.

- encode a seven-bit mark epoch plus current gray bit in the existing byte per handle
- advance the epoch in O(1) at cycle start instead of whitening every live handle
- retain black survivor state during sweep instead of rewriting each survivor white
- initialize new and reused handles with phase-correct current-epoch state
- preserve synchronous `CollectFull` restart semantics, including a restart across epoch wrap
- strengthen verification, telemetry restart coverage, fuzzing, and footprint controls

## Correctness invariants

- a mark state from any non-current epoch is logically white
- current-epoch high-bit state is gray; current-epoch low state is black
- active-cycle restart advances to a third epoch, distinct from both current marks and the preceding white population
- every live object is black in the current epoch whenever Tiny is idle
- gray objects remain represented exactly once by the queue or active partial-scan cursor
- new pointerful mark/remark allocations are gray; idle and sweep-protected allocations are black

## Measurements

Five one-second `GOMAXPROCS=1` cycle-start samples on the Ryzen 7 8845HS:

| Handles | `main` | Current | Change |
|---:|---:|---:|---:|
| 256 | 237.5 ns | 4.341 ns | -98.17% |
| 4,096 | 3.698 us | 4.223 ns | -99.89% |
| 65,536 | 60.959 us | 4.510 ns | -99.99% |

All remain 0 B/op and 0 allocs/op. Ten CPU-affined, single-P, interleaved 300 ms controls at 4,096 reference-array elements measured +1.68% for an individual bounded Step and +0.30% for a complete cycle; median paired deltas were -0.13%/+0.19%.

Metadata remains one byte per handle, `tinyGC` remains 80 bytes, and `Collector` remains 1,120 bytes on 64-bit. The final release size card reports no aligned-file change for runtime-standard, runtime-minimal -4.0 KiB, and runtime-minimal-tiny +0.1 KiB versus `main`; every profile remains within budget.

## Verification

- `go test ./... -count=1`
- `go test ./src/... -count=1`
- `go test -tags wago_gcstats ./src/core/runtime/gc ./src/wago -count=1`
- `wagodebug`, `wagodebug+wago_gcstats`, `wago_guardpage`, and `wago_guardpage+wago_gcstats` matrices
- `go test -race ./src/core/runtime/gc -count=1`
- checkptr and `go vet`
- Linux/ARM64 collector test compile
- stateful Tiny fuzz smoke: 476,353 ordinary and 172,790 telemetry-tagged executions
- Tiny allocation-bound fuzz smoke: 628,604 executions

## Scope boundary

This PR intentionally leaves resumable root enumeration, bounded sweep/allocation, allocation-debt pacing, bounded barrier drains, SATB comparison, and Tiny product splitting for later #319 stages.


jtenner:
@codex review

jtenner:
Adversarial review follow-up pushed in `59d99917`:

- removed Tiny epoch encoding from the shared `newHandle` append path; the byte is unpublished there and Tiny's mandatory post-allocation hook already writes the authoritative current-epoch gray/black state, so Throughput handle growth no longer performs irrelevant epoch work
- added a wrap-boundary synchronous restart test from `tinySweep` after sweep has already advanced past a black survivor, proving the fresh epoch reconsiders that survivor and preserves only the new exact roots

Repeated focused tests, full `src/...`, telemetry/debug/guard-page tags, race, checkptr, vet, and another stateful Tiny fuzz smoke all pass.

jtenner:
Follow-up measurement for the shared-handle review fix: a CPU-affined, pre-grown 100,000-handle microbenchmark measured median handle publication at 4.939 ns on `main`, 6.745 ns in the pre-fix epoch revision (+36.6%), and 4.879 ns after `59d99917` (-1.2% versus `main`), all at 0 B/op and 0 allocs/op. The temporary review benchmark/worktrees were removed.

jtenner:
@codex review

jtenner:
Adversarial review completed with two additional fixes in `83d8fd84` and `e05c0e98`:

1. **Hot-path regression:** the initial epoch implementation imposed generic color encoding and inlined cold white-child shading into the active Tiny barrier. Branch-free current-epoch setters, proven-color queueing, and a cold noinline white-child helper restore the checked marked-child barrier to within +1.26% of `main` (paired median +2.34%), 0 B/op and 0 allocs/op. A permanent active Tiny barrier benchmark now guards this path.
2. **Sweep root-publication corruption:** extended stateful fuzzing found a checked global store could publish a white pointerful graph after an earlier descendant had already been swept. Checked global/table creation and setters now reject that unsafe publication before mutating the slot. Marked graphs remain accepted, and white pointer-free objects retain immediate sweep marking. The minimized reproducer is committed as `c7f7ce9b775e073a`.

The final size card is within budget: runtime-standard unchanged at file granularity, runtime-minimal -4.0 KiB, and runtime-minimal-tiny +0.1 KiB versus `main`. Final validation passed full `src/...`, telemetry/debug/guard-page tags, race, checkptr, vet, arm64 compile, 1,515,770 ordinary stateful fuzz executions, 578,654 telemetry-tagged executions, and 751,594 allocation-bound executions.

jtenner:
@codex review

PR #423, opened 2026-08-12
cli: harden release downloads and credential persistence

### LINKED ISSUE #402: [P0][CLI] Add context deadlines and bounded response bodies to network clients
## Problem

Several CLI network paths use `http.Get`, `http.PostForm`, or `http.DefaultClient.Do` without a request context or client deadline. Several of those paths also read response bodies to EOF without a size bound.

A server, proxy, captive portal, or test endpoint can therefore hold commands indefinitely or stream unbounded data into memory. This affects update discovery/downloads, registry operations, and OAuth/device-flow requests.

## Evidence

- [Release downloads use `http.Get` and buffer until EOF](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/download.go).
- [Release/channel discovery uses the default HTTP path](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/discovery.go).
- [Registry API requests use `http.DefaultClient.Do` and unbounded `io.ReadAll`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/registry/client.go).
- [OAuth/device requests use default-client form posts](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/registry/oauth.go).
- [Authentication/status paths call the same unbounded helpers](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/registry/auth.go).

`recordRegistryInstall` already uses a two-second request context, demonstrating the intended bounded pattern, but the primary authenticated operations do not.

## Impact

- `wago update`, version selection, login, registry lookup, publish/install, or account commands can hang forever.
- cancellation from the command/process cannot reliably interrupt DNS, connect, TLS, response-header, or body reads.
- small JSON/checksum endpoints can cause unbounded memory growth.
- error responses may be arbitrarily large and then be retained solely to display a short message.
- tests that use a stalled server can leak goroutines or block the suite.

## Suggested fix

Create shared, injectable HTTP clients and thread command contexts through every request.

Recommended controls:

- transport-level dial, TLS-handshake, idle-connection, and response-header timeouts;
- operation-specific overall context deadlines;
- `http.NewRequestWithContext` for every request;
- bounded response readers with endpoint-specific limits, e.g. small limits for checksums/OAuth/API JSON and a separately bounded streaming path for release assets;
- reject a declared `Content-Length` above the endpoint limit before reading;
- when length is unknown, use `io.LimitReader(limit+1)` and return an explicit oversized-response error;
- bound error-body capture independently (for example 64 KiB) and drain/close safely where connection reuse matters;
- expose clients/timeouts in internal configuration for deterministic tests, without allowing insecure production defaults accidentally.

Avoid one arbitrary global timeout for all operations: release asset transfer may need a longer total deadline than a small registry request, while both still need progress/read-stall cancellation.

## Regression tests

Use `httptest.Server` handlers that:

- never send headers;
- send headers and then stall the body;
- stream endless chunked data;
- declare an oversized `Content-Length`;
- return a huge JSON/error body;
- cancel the parent context mid-request.

Verify bounded return time, bounded allocation, response-body closure, and actionable errors for each command family.

## Acceptance criteria

- no production network path uses the process-default client without an explicit context/deadline policy.
- all in-memory response bodies have documented maximum sizes.
- command cancellation interrupts active network I/O.
- release downloads use their own streaming/bounded implementation rather than the JSON helper.
- tests prove no indefinite wait or unbounded read for every endpoint class.

## Audit context

### LINKED ISSUE #403: [P0][Updater] Stream release assets through a bounded hash-verifying temp file
## Problem

The version manager downloads each executable release asset into a `bytes.Buffer`, fetches the checksum afterward, hashes the in-memory byte slice, and only then writes the complete binary to disk.

This requires memory proportional to the release asset size and has no maximum transfer bound. `WAGO_RELEASE_BASE` is configurable, so a compromised/misconfigured endpoint or proxy can cause the manager to consume arbitrary RAM before verification.

## Evidence

- [`downloadReleaseAssetWithProgress` calls `httpGetBytesProgress` for the executable and retains the returned `[]byte`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/download.go).
- [`httpGetBytesProgress` copies every response chunk into an unbounded `bytes.Buffer`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/download.go).
- checksum verification happens only after the entire body is resident, followed by a second full copy into `os.WriteFile`.

## Impact

- peak memory is at least the full binary size plus buffer growth/copy overhead.
- an endless or oversized response can exhaust memory before checksum verification.
- large builds create avoidable GC pressure and transient RSS spikes during self-update/plugin runtime installation.
- partial downloads cannot be resumed and may leave the fixed `dest + ".tmp"` path behind on failure.
- simultaneous downloads targeting the same destination can collide on the fixed temporary filename.

## Suggested fix

Use a same-directory unique temporary file and hash while streaming:

1. fetch and strictly parse the small checksum file first, with a tiny body limit;
2. create a unique temp file via `os.CreateTemp(filepath.Dir(dest), ...)` with mode `0600`/`0700` during construction;
3. stream the asset through `io.MultiWriter(temp, sha256.New())` with a maximum-byte counter;
4. reject an oversized declared `Content-Length` before transfer and reject unknown-length transfers once the configured cap is exceeded;
5. update progress from streamed bytes without retaining them;
6. flush/sync and close the file;
7. compare the digest using a strict 64-hex parser (and expected filename when checksum format provides one);
8. set final executable permissions and install through the platform-aware replacement helper;
9. remove the temp file on every failure/cancellation path.

The maximum can be a conservative constant well above current release sizes or, preferably, release metadata signed/published with the expected size. It must not be inferred solely from untrusted `Content-Length`.

## Security details

- do not write the final destination before checksum success.
- prevent symlink/path substitution at the destination and temp-file boundary.
- avoid following redirects to unsupported schemes/hosts under restrictive update policy.
- use constant-time digest comparison where practical, though digest secrecy is not required.
- reject malformed, truncated, or multi-hash checksum payloads deterministically.

## Regression tests

- large valid asset: memory remains bounded and digest passes.
- unknown-length endless stream: transfer stops at the cap and temp file is removed.
- oversized `Content-Length`: reject before reading the body.
- checksum mismatch/truncated response/cancellation: destination remains unchanged.
- concurrent downloads use distinct temp files.
- successful install preserves executable mode and replaces the prior version on every supported OS.

## Acceptance criteria

- updater memory usage is O(buffer size), not O(asset size).
- every asset transfer has an explicit maximum.
- bytes are hashed during the single disk write; no full in-memory copy exists.
- the existing destination changes only after complete verified transfer.
- temporary artifacts are unique and cleaned on all failures.

## Audit context

Observed on `main` at `ca3266ca7adca444f17d2f1f225d71854bba8ad7`. Related network timeout/body-limit work: #402.

### LINKED ISSUE #404: [P0][Cross-platform] Centralize replace-existing semantics for updater and artifact cache
## Problem

Several install/cache paths implement “atomic replacement” as `os.Rename(temp, destination)` while the destination may already exist. That works as replacement on Unix-like systems but does not provide replace-existing behavior on Windows.

The repository already contains Windows-specific replacement logic for self-management, but the release updater, source-built runtime installer, and artifact cache bypass it.

## Evidence

- [Downloaded release installation writes `dest + ".tmp"` and calls `os.Rename(tmp, dest)`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/download.go).
- [Source-built runtime/manager installation calls `os.Rename(tempRunner, dest)`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/version/source.go).
- [Artifact-cache writes end with `os.Rename(tempName, path)`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/runtime/internal/artifactcache/cache.go).
- [`LoadOrCompile` discards cache-write errors, so a failed replacement is silent](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/runtime/internal/artifactcache/cache.go).
- [The self-replacement package already uses Windows `MoveFileExW` with `MOVEFILE_REPLACE_EXISTING` where replacement is required](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/self/replace/replace_windows.go).
- [Artifact-cache documentation describes writes as atomic replacement](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/docs/artifact-cache.md).

## Impact

- updating an already-installed binary can fail on Windows even after a valid download/build.
- an existing corrupt artifact cache entry can become effectively permanent: every execution decodes it, recompiles, fails to replace it, and silently repeats the cycle.
- successful source rebuilds can fail at the final install step.
- fixed temp names such as `dest + ".tmp"` can collide across concurrent operations and can be left behind.
- platform-specific behavior is duplicated and inconsistently tested.

## Suggested fix

Create one internal `atomicfile`/`replacefile` abstraction with explicit semantics:

```go
ReplaceFile(destination string, mode fs.FileMode, write func(io.Writer) error) error
```

It should:

- create a unique same-directory temporary file;
- set restrictive construction permissions and final mode deliberately;
- write, flush, optionally `fsync`, and close before replacement;
- replace an existing regular file on every supported OS using platform-specific primitives;
- reject or safely handle symlink/directory destinations;
- clean temporary files on all failures;
- optionally sync the parent directory on platforms where durability is promised;
- surface errors to callers instead of silently discarding them.

On Windows, use `ReplaceFileW`/`MoveFileExW(MOVEFILE_REPLACE_EXISTING | MOVEFILE_WRITE_THROUGH)` as appropriate, with sharing-violation behavior documented. Avoid a remove-then-rename sequence unless the brief missing-destination window is explicitly acceptable; that is not atomic.

The updater may still require its dedicated running-executable replacement flow, but cache/config/data files should share one tested helper.

## Regression tests

- replace an existing destination on Windows and Unix.
- repair an existing corrupt cache entry and verify the next run hits it.
- destination is a directory/symlink/read-only/open file.
- two concurrent writers do not share a temp path or leave debris.
- injected write/close/sync/replace failures preserve the old destination.
- successful replacement has the intended final permissions.

## Acceptance criteria

- update/source-build/cache paths use one platform-correct replace-existing primitive.
- cache write failures are observable or logged with enough detail to diagnose persistent misses.
- an old valid destination survives every pre-commit failure.
- Windows integration tests cover replacement of an existing binary and cache artifact.
- documentation matches the actual durability/atomicity guarantees.

## Audit context

### LINKED ISSUE #408: [P0][Security] Atomically rewrite registry credentials and enforce 0600
## Problem

Registry bearer tokens are stored by directly calling `os.WriteFile(path, data, 0o600)` after a read-modify-write cycle.

This has three security/correctness problems:

1. the permission argument only controls creation; an existing credentials file with permissive mode is not tightened to `0600`;
2. direct truncation/rewrite can leave an empty or partial JSON file after a crash, disk-full condition, or interrupted write;
3. concurrent login/logout/process operations can read the same old map and overwrite each other's changes.

## Evidence

- [`writeCredentials` creates the directory, marshals JSON, and calls `os.WriteFile` directly](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/registry/config.go).
- [`saveCredentials` and `deleteCredentials` each perform separate load-modify-write sequences without a lock](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/registry/config.go).
- credentials contain long-lived bearer tokens in plaintext.

## Impact

- a pre-existing `0644`/group-readable file remains readable after Wago “writes with 0600.”
- process interruption can corrupt all stored registry credentials, including unrelated registries.
- simultaneous operations can lose a newly stored token or resurrect a deleted one.
- readers can observe partial JSON while another process truncates and rewrites.
- symlink/path replacement around the credentials file is not handled explicitly.

## Suggested fix

Use a serialized, atomic credential-store transaction:

1. create/open a project-independent lock file in the config directory and acquire a cross-process exclusive lock;
2. re-read the latest credentials only after obtaining the lock;
3. apply the mutation;
4. create a unique same-directory temporary file with `0600` from the first open;
5. write JSON, flush, `fsync`, close, and explicitly `chmod(0600)` as a defense against inherited/umask/platform surprises;
6. replace the destination using the platform-aware helper from #404;
7. sync the parent directory if crash durability is part of the contract;
8. release the lock and clean every temporary file on failure.

Before replacing, reject an unexpected directory/symlink destination or use no-follow/open-by-handle semantics appropriate to each platform. Also verify or repair the config directory's restrictive mode where supported.

Consider OS credential-vault integration as a separate enhancement. Even with a keychain, the fallback file path must remain atomic and permission-safe.

## Regression tests

- start with an existing credentials file at `0644`; after a write it is `0600` on permission-bearing systems.
- inject short write, close, sync, and replace failures; the old valid file remains intact.
- concurrent saves for two registries preserve both entries.
- concurrent save/delete has a deterministic serialized result and never malformed JSON.
- readers never observe a truncated intermediate file.
- symlink/directory destination is rejected safely.
- Windows replacement and ACL behavior is documented/tested separately from Unix mode bits.

## Acceptance criteria

- successful writes leave a complete, parseable credential set with restrictive permissions.
- an interrupted/failed update preserves the prior valid file.
- concurrent CLI processes cannot lose unrelated registry entries.
- existing permissive files are repaired, not merely rewritten with a creation-mode argument.
- every token-bearing temporary file is restrictive and removed after failure.

## Audit context

Observed on `main` at `ca3266ca7adca444f17d2f1f225d71854bba8ad7`. Related replacement primitive: #404.

jtenner:
## Summary

- add shared bounded, context-aware HTTP policies for release, registry, and OAuth operations
- stream release assets through fixed-size buffers, verify strict SHA-256 records, and publish only complete verified executables
- centralize same-directory atomic replace-existing behavior across Unix and Windows for updater, source-build, plugin-build, cache, and credential paths
- serialize registry credential mutations across processes and publish private `0600` files atomically
- propagate cancellation through source Git/Go/TinyGo subprocesses and bound aggregate release discovery and OAuth device-flow waits

## Safety properties

- stalled requests have explicit connection, header, and overall deadlines
- in-memory response bodies and streamed transfers have endpoint-specific limits
- failed or canceled updates preserve the previously installed executable
- concurrent writers use unique temporary files
- Windows replacement uses `MoveFileExW` replace-existing semantics
- credential save/logout decisions happen while holding an OS-backed cross-process lock
- malformed, null, directory, and symlink credential destinations are rejected without overwriting prior contents

## Review fixes

A second-pass audit found and corrected several edge cases before opening this PR:

- release-generated checksum files used `./asset`, which the initial strict parser rejected
- self-update still used a fixed `.new` staging path
- source-build subprocesses ignored command cancellation
- repeated full metadata pages and server-provided OAuth timing values could extend work excessively
- oversized streams wrote one detection byte beyond their documented disk cap
- logout checked credential existence before acquiring the transaction lock
- a JSON `null` credential store could produce a nil-map panic

## Validation

- `go test ./...`
- `go test -race ./...`
- `go vet ./...`
- `tinygo test ./cli/runtime/internal/artifactcache`
- repeated focused updater, registry, atomic-file, file-lock, and self-update tests
- `git diff --check`
- shell syntax and checksum-format checks for release scripts
- production builds for Linux amd64/arm64, Darwin amd64/arm64, and Windows amd64/arm64
- affected package test cross-compilation for Windows amd64/arm64

## Memory

For a 16 MiB release asset, the streamed implementation allocates roughly 100–160 KiB per operation instead of approximately 42 MiB. Allocation volume remains approximately flat as asset size increases.

Closes #402
Closes #403
Closes #404
Closes #408

### DIFF
--- a/cli/manager/command_environment.go
+++ b/cli/manager/command_environment.go
@@ -1,6 +1,7 @@
 package manager
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -43,7 +44,14 @@ import (
 
 // commandEnvironment adapts manager-owned domain modules to command packages.
 // Parsing, validation, and command-level orchestration stay in command.go.
-type commandEnvironment struct{}
+type commandEnvironment struct{ ctx context.Context }
+
+func (e commandEnvironment) context() context.Context {
+	if e.ctx != nil {
+		return e.ctx
+	}
+	return context.Background()
+}
 
 func (commandEnvironment) Compile(options compilecmd.Options) {
 	if err := managerplugin.Select(options.Global, options.Local, options.Bare); err != nil {
@@ -304,7 +312,7 @@ func (e commandEnvironment) CachePrune(options cacheprune.Options) {
 	fmt.Printf("Pruned %d old cache %s (%s)\n", result.Removed, noun, managercache.FormatBytes(result.Bytes))
 }
 
-func (commandEnvironment) Login(options authlogin.Options) {
+func (e commandEnvironment) Login(options authlogin.Options) {
 	if automation.DryRun() {
 		method := "interactive"
 		switch {
@@ -320,28 +328,28 @@ func (commandEnvironment) Login(options authlogin.Options) {
 		automation.PrintPlan("log in", map[string]any{"registry": "plugins.wago.sh", "method": method})
 		return
 	}
-	registry.Login(registry.LoginRequest{
+	registry.LoginContext(e.context(), registry.LoginRequest{
 		Link: options.Link, Code: options.Code, WithToken: options.WithToken, Token: options.Token,
 	})
 }
 
-func (commandEnvironment) Logout() {
+func (e commandEnvironment) Logout() {
 	if automation.DryRun() {
 		automation.PrintPlan("log out", map[string]any{"registry": "plugins.wago.sh"})
 		return
 	}
-	registry.Logout()
+	registry.LogoutContext(e.context())
 }
-func (commandEnvironment) Whoami() { registry.Whoami() }
+func (e commandEnvironment) Whoami() { registry.WhoamiContext(e.context()) }
 
-func (commandEnvironment) Add(options pluginadd.Options) {
+func (e commandEnvironment) Add(options pluginadd.Options) {
 	requireUnlocked("add plugins")
 	if automation.DryRun() {
 		automation.PrintPlan("add plugins", map[string]any{"packages": options.Modules, "scope": scopeName(options.Global, options.Local), "force": options.Force})
 		return
 	}
 	managerplugin.Add(managerplugin.AddRequest{
-		Modules: options.Modules, Global: options.Global, Local: options.Local,
+		Context: e.context(), Modules: options.Modules, Global: options.Global, Local: options.Local,
 		Force: options.Force, Verbose: options.Verbose,
 		Capabilities: options.Capabilities, GrantAll: options.GrantAll, DenyAll: options.DenyAll,
 	})
@@ -368,14 +376,14 @@ func (commandEnvironment) Grant(options grant.Options) {
 	})
 }
 
-func (commandEnvironment) UpdatePlugins(options pluginupdate.Options) {
+func (e commandEnvironment) UpdatePlugins(options pluginupdate.Options) {
 	requireUnlocked("update plugins")
 	if automation.DryRun() {
 		automation.PrintPlan("update plugins", map[string]any{"package": options.Module, "scope": scopeName(options.Global, options.Local), "force": options.Force})
 		return
 	}
 	managerplugin.Update(managerplugin.MutationRequest{
-		Name: options.Module, Global: options.Global, Local: options.Local,
+		Context: e.context(), Name: options.Module, Global: options.Global, Local: options.Local,
 		Force: options.Force, Verbose: options.Verbose,
 	})
 }
@@ -397,31 +405,31 @@ func (commandEnvironment) RebuildPlugins(options pluginrebuild.Options) {
 	managerplugin.RebuildRuntime(maintenanceRequest("", options.Global, options.Local, options.Verbose))
 }
 
-func (commandEnvironment) Publish(options publish.Options) {
+func (e commandEnvironment) Publish(options publish.Options) {
 	if automation.DryRun() {
 		automation.PrintPlan("publish plugin", map[string]any{"manifest": options.Manifest, "commit": options.Commit, "category": options.Category, "tags": options.Tags})
 		return
 	}
-	registry.Publish(registry.PublishRequest{
+	registry.PublishContext(e.context(), registry.PublishRequest{
 		Manifest: options.Manifest, Commit: options.Commit, Notes: options.Notes,
 		Category: options.Category, Tags: options.Tags,
 	})
 }
 
-func (commandEnvironment) Unpublish(options unpublish.Options) {
+func (e commandEnvironment) Unpublish(options unpublish.Options) {
 	if automation.DryRun() {
 		automation.PrintPlan("unpublish plugin", map[string]any{"target": options.Target})
 		return
 	}
-	registry.Unpublish(registry.UnpublishRequest{Target: options.Target, Yes: options.Yes})
+	registry.UnpublishContext(e.context(), registry.UnpublishRequest{Target: options.Target, Yes: options.Yes})
 }
 
-func (commandEnvironment) Deprecate(options deprecate.Options) {
+func (e commandEnvironment) Deprecate(options deprecate.Options) {
 	if automation.DryRun() {
 		automation.PrintPlan("change plugin deprecation", map[string]any{"target": options.Target, "message": options.Message, "undo": options.Undo})
 		return
 	}
-	registry.Deprecate(registry.DeprecateRequest{
+	registry.DeprecateContext(e.context(), registry.DeprecateRequest{
 		Target: options.Target, Message: options.Message, Undo: options.Undo,
 	})
 }
@@ -443,7 +451,7 @@ func (commandEnvironment) dirs() wagopaths.Dirs {
 }
 
 func (e commandEnvironment) toolchain() managerversion.Toolchain {
-	return managerversion.Toolchain{Dirs: e.dirs()}
+	return managerversion.Toolchain{Dirs: e.dirs(), Context: e.context()}
 }
 
 func (e commandEnvironment) List()    { e.toolchain().List() }
@@ -491,12 +499,12 @@ func (e commandEnvironment) UninstallAllVersions() {
 	e.toolchain().UninstallAll()
 }
 
-func (commandEnvironment) Update(force bool) {
+func (e commandEnvironment) Update(force bool) {
 	if automation.DryRun() {
 		automation.PrintPlan("update Wago", map[string]any{"component": "manager", "force": force})
 		return
 	}
-	managerself.Update(versionString(), managerself.ExecutablePath(), force)
+	managerself.UpdateContext(e.context(), versionString(), managerself.ExecutablePath(), force)
 }
 
 func (e commandEnvironment) UpdateEverything(options updatecmd.Options) {
@@ -509,7 +517,7 @@ func (e commandEnvironment) UpdateEverything(options updatecmd.Options) {
 	}
 	activeRuntime := managerversion.ActiveVersion(e.dirs())
 	if options.Manager {
-		managerself.Update(versionString(), managerself.ExecutablePath(), options.Force)
+		managerself.UpdateContext(e.context(), versionString(), managerself.ExecutablePath(), options.Force)
 	}
 	if options.Runtime {
 		channel := options.Channel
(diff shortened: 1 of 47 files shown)


jtenner:
CI is green on commit `6922c66f`. The follow-up fixes address Windows sharing conflicts, staticcheck failures, and the Ctrl-C review finding. Auto-merge is enabled and is waiting only for the required review from @JairusSW.

PR #425, opened 2026-08-12
runtime: complete lifecycle and cache hardening

### LINKED ISSUE #388: [P0][Runtime] Reject compilation after Runtime.Close
## Problem

`Runtime.CloseContext` marks the runtime closed, and most runtime-bound APIs reject further work. `Runtime.Compile` is the exception: it snapshots configuration/hooks/instructions and performs native compilation without checking `rt.closed`.

That permits expensive code generation and extension hook execution after the runtime has been torn down. The returned `Module` is bound to the already-closed runtime and cannot be instantiated successfully, so the work is unusable. A concurrent `Compile`/`Close` pair also has undefined ordering from the caller's perspective.

## Evidence

- [`Runtime.CloseContext` records the closed state](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).
- [`Runtime.Module`, `Runtime.Instantiate`, `Runtime.Use`, and `Runtime.LoadPlugins` check the closed state](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).
- [`Runtime.Compile` does not perform the equivalent check](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).

A minimal regression test should cover:

```go
rt := NewRuntime()
_ = rt.Close()
_, err := rt.Compile(validModule)
// err should identify a closed runtime before hooks or native compilation run.
```

## Impact

- Native executable mappings and compiler allocations can be created after shutdown.
- compile hooks may observe or mutate a runtime whose services/reference store are already closed.
- callers receive a module that is structurally valid but unusable with its owning runtime.
- concurrent shutdown semantics differ from the other runtime-bound operations.

## Suggested fix

Check `rt.closed` while holding `rt.mu` at the same point where compile configuration, hooks, imports, and instruction bindings are snapshotted. Return a stable closed-runtime error before invoking any extension callback or compiler work.

Do not use an unlocked pre-check; it would leave a `Compile`/`Close` TOCTOU window. Define the lock acquisition as the operation's admission point, then run callbacks/compiler work after releasing the mutex.

## Acceptance criteria

- serial compilation after `Close` fails before any compile hook or compiler invocation.
- a compile admitted before shutdown may complete under explicitly documented semantics; a compile arriving after shutdown admission fails deterministically.
- race tests cover `Compile` against `CloseContext`.
- the result is consistent with `Module`, `Instantiate`, `Use`, and `LoadPlugins`.

## Audit context

### LINKED ISSUE #390: [P0][Runtime] Isolate shutdown callback panics and guarantee final store teardown
## Problem

`Runtime.CloseContext` invokes plugin `Stop` methods, internal service close functions, and `OnRuntimeClose` hooks directly. A panic from any one callback aborts the remainder of shutdown and can skip `referenceStore.closeRuntime`.

Other lifecycle paths already use panic-isolation helpers for extension callbacks, so shutdown currently has weaker containment exactly where best-effort cleanup is most important.

## Evidence

- [`Runtime.CloseContext` iterates plugin stops, internal closes, and runtime-close hooks without recovery](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).
- [`callHookSafely` exists for lifecycle-hook panic isolation](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/hooks.go).
- the final reference-store shutdown occurs only after the callback loops in [`Runtime.CloseContext`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).

## Impact

- later plugins/services never receive cleanup.
- reference tokens, GC domains, retained instances, and native mappings can remain live.
- callers see a process panic instead of the documented aggregated shutdown error behavior.
- repeated `Close` cannot necessarily repair the damage because the runtime is already marked closed.

## Suggested fix

Treat every extension-controlled shutdown callback as an independent cleanup step:

- wrap plugin `Stop`, internal close, and runtime-close hook calls with panic recovery that converts the panic to an attributed error;
- continue running all remaining callbacks;
- guarantee the reference-store close in a final/deferred phase even when callback code fails;
- preserve reverse plugin order and current error joining;
- ensure a panic value/stack is bounded and does not leak secrets through normal errors.

The store close itself should remain outside extension callbacks and be attempted exactly once.

## Acceptance criteria

- a panic in the first plugin `Stop` does not prevent later stops, internal closes, runtime hooks, or store teardown.
- panics are returned as attributed errors rather than escaping `CloseContext`.
- multiple panics/errors are aggregated deterministically.
- close remains idempotent.
- tests cover panics in every callback category and verify retained instance/token resources are released.

## Audit context

### LINKED ISSUE #393: [P0][Runtime] Reject Modules bound by a different Runtime
## Problem

`Module` records the `Runtime` that compiled or bound it, but `Runtime.Instantiate` never verifies that the module belongs to the receiver.

A module bound by `rtA` can therefore be passed to `rtB.Instantiate`. The operation applies capability requirements and structured import metadata derived from `rtA`, then merges imports and executes hooks/reference-store behavior from `rtB`.

This crosses a security and ownership boundary: one runtime's policy view is being executed under another runtime's import namespace and lifecycle.

## Evidence

- [`Module` has an owning `rt *Runtime` field and is documented as runtime-aware](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/module.go).
- [`Runtime.buildModule` derives `Provided`, capability, documentation, and required-capability metadata from that runtime](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/module.go).
- [`Runtime.instantiateOrigin` applies policy to `mod`, then uses the receiver's imports, hooks, and reference store without checking `mod.rt == rt`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).

## Concrete policy mismatch

1. Bind a compiled module under `rtA`, where an imported function has no capability metadata.
2. Configure `rtB` with a provider for the same import that declares a privileged capability.
3. Call `rtB.Instantiate(ctx, moduleFromA, WithPolicy(restrictivePolicy))`.
4. `applyPolicy` sees `moduleFromA.reqCaps`, not the requirements derived from `rtB`'s provider.
5. binding and execution proceed against `rtB`'s privileged import namespace.

Even without a capability mismatch, `Provided` flags, docs, import names, compile hooks, and module lifecycle context all refer to the wrong runtime generation.

## Impact

- capability policy can be evaluated against stale or unrelated metadata.
- extension import/provider changes in the destination runtime are not reflected in the module.
- compile/instantiate hooks receive a `Module` whose `Runtime` identity disagrees with the operation.
- GC/reference-store ownership and future module-owned extension state become harder to reason about.

## Suggested fix

Reject cross-runtime modules at the beginning of `Runtime.instantiateOrigin`:

```go
if mod.rt != rt {
    return nil, ErrForeignModule // stable/wrappable error
}
```

Callers that intentionally want to reuse compiled code should explicitly rebind it:

```go
modB, err := rtB.Module(modA.Compiled())
```

That recomputes import metadata, capability requirements, and lifecycle hooks for the destination runtime while preserving compiled-code reuse.

If modules may outlive or move between runtime state generations in the future, store and validate an immutable runtime/generation identity rather than only a pointer.

## Acceptance criteria

- `rtB.Instantiate` rejects a `Module` created by `rtA` before policy checks, import lookup, hooks, or native instantiation.
- explicit `rtB.Module(compiled)` rebinding remains supported.
- tests cover mismatched capability metadata, identical imports on two runtimes, closed source/destination runtimes, and concurrent use.
- the error is stable enough for callers to identify and remediate.

## Audit context

### LINKED ISSUE #400: [P0][Cache] Propagate Runtime.Module errors instead of treating them as artifact misses
## Problem

The artifact cache treats every failure after reading a cache file as a cache miss. In particular, after a cached `Compiled` artifact successfully decodes, an error from `rt.Module(compiled)` is silently discarded and the source is compiled again.

`Runtime.Module` is not merely cache validation. It binds the artifact to the runtime and runs `AfterCompile` hooks. Those hooks may implement policy, authorization, provenance checks, accounting, or extension-owned metadata. Swallowing their errors can execute the same policy twice and changes semantics between cache-hit and source-compile paths.

The successfully decoded `Compiled` object is also not closed when binding fails, leaving its executable mapping to finalization.

## Evidence

- [`Cache.LoadOrCompile` ignores `rt.Module` errors and falls through to `rt.Compile`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/runtime/internal/artifactcache/cache.go).
- [`Runtime.Module` rejects closed runtimes, builds runtime-specific metadata, and executes `AfterCompile` hooks](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).
- [`Compiled.Close` is required for deterministic code-image cleanup](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/code_mapping.go).

## Problematic cases

- a closed runtime produces a `Runtime.Module` error, but the cache then spends time compiling before returning a different error.
- an `AfterCompile` hook denies artifact use; the cache recompiles and invokes it again, potentially succeeding if the hook is stateful or one-shot.
- an accounting/audit hook records duplicate attempts.
- a transient hook failure causes needless full compilation rather than surfacing the actual runtime-binding failure.
- repeated failures retain decoded native mappings until GC.

## Suggested fix

Separate cache corruption/incompatibility from runtime-binding failure:

1. read/decode failures that prove the artifact is absent, corrupt, or incompatible may be treated as a miss (and ideally remove/quarantine the bad entry);
2. once `UnmarshalBinary` succeeds, call `rt.Module` exactly once;
3. if binding fails, close the decoded `Compiled` and return the binding error;
4. only compile source when the artifact itself could not be used before runtime policy/lifecycle binding began.

If some `Runtime.Module` errors are explicitly retryable through recompilation, introduce typed error classification and document why recompiling can change the result. Do not use a blanket `err == nil` test.

This should coordinate with deterministic module ownership in #394.

## Regression tests

- cached artifact plus a rejecting `AfterCompile` hook: hook runs once, source compiler is not invoked, mapping is released.
- cached artifact plus closed runtime: return the closed-runtime error without compilation.
- corrupt/incompatible artifact: compile source once and repair the cache.
- a stateful hook that would allow a second invocation must not bypass its first denial.
- instrument code-cache references to prove no finalizer-dependent cleanup on failed binding.

## Acceptance criteria

- runtime binding/policy errors are never silently converted into cache misses.
- successful artifact decode followed by failure has deterministic cleanup.
- cache corruption remains best-effort and does not prevent safe recompilation.
- cache-hit and cold-compile paths invoke runtime lifecycle policy with equivalent exactly-once semantics.

## Audit context

### LINKED ISSUE #401: [P2][Cache] Remove FunctionWorkers from artifact compatibility identity
## Problem

The automatic artifact cache includes `RuntimeConfig.FunctionWorkers()` in its compatibility key even though worker count is a transient scheduling policy and the compiler documentation explicitly guarantees that generated output and serialized `.wago` artifacts are independent of worker count.

Changing from serial to adaptive or changing a worker limit therefore forces unnecessary cold compilation and creates duplicate cache entries for byte-identical artifacts.

## Evidence

- [`artifactcache.Cache.path` appends `config.FunctionWorkers()` to the hashed cache signature](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/runtime/internal/artifactcache/cache.go).
- [The function-worker documentation states that functions are joined in original order and generated output/serialized artifacts are independent of worker count](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/docs/function-workers.md).
- `Compiled` intentionally retains no worker policy according to the same document.

## Impact

- `-p`, `-p4`, adaptive mode, and default serial mode cannot share otherwise compatible artifacts.
- the same source/config/build can occupy multiple cache files solely because of machine CPU count or an operational tuning change.
- deployment changes intended only to alter cold compile latency unexpectedly discard warm-cache behavior.
- cache hit-rate measurements are polluted by a non-semantic key dimension.

## Suggested fix

Remove `FunctionWorkers` from artifact identity and bump the cache-key format so old entries remain safely isolated.

Add a classification rule to the cache-key design:

- include settings that change decoded semantics, validation admission, emitted machine code, native ABI, serialized metadata, bounds behavior, or runtime interpretation;
- exclude settings that affect only scheduling, transient memory, logging, progress, or compiler parallelism.

Coordinate this with the broader generated-fingerprint work in #328, which should preserve the same semantic/non-semantic distinction.

## Regression tests

For a representative corpus, compile with worker policies `0`, `1`, `2`, `4`, and a value above `GOMAXPROCS` and assert:

- identical `Compiled.MarshalBinary()` bytes;
- identical native function bytes/metadata where exposed;
- one cache path/key after this fix;
- a warm hit when a second process changes only worker policy.

Retain existing determinism tests for serial versus parallel validation/code generation. If any module is found whose output differs by worker count, that is a compiler determinism defect and should be fixed rather than encoded as cache compatibility.

## Acceptance criteria

- worker-policy changes reuse compatible artifacts.
- settings that genuinely alter code or artifact interpretation remain in the key.
- cache-key format/version is bumped deliberately.
- documentation lists semantic key dimensions and excluded scheduling-only options.

## Audit context

### LINKED ISSUE #416: [P1][API] Deep-copy nested inspection data
## Problem

Two public inspection APIs return only shallow copies of structs that contain slices/maps:

- `Module.Imports()` copies the outer `[]ImportSpec`, but each returned `ImportSpec` still aliases the module's internal `Params`, `Results`, `ParamTypes`, and `ResultTypes` backing arrays.
- `Runtime.Extensions()` copies the outer `[]ExtensionInfo`, but nested provenance, ordering, capability, and compatibility containers still alias runtime state.

`Runtime.Use` also stores the `ExtensionInfo` returned by plugin code without first deep-copying it, so a plugin can mutate backing slices/maps after successful registration.

The APIs look like immutable snapshots, but callers can modify the runtime/module's retained metadata and race with concurrent readers.

## Evidence

- [`ImportSpec` contains four slice fields](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/module.go).
- [`Module.Imports` is implemented as `append([]ImportSpec(nil), m.imports...)`, which clones only the outer slice](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/module.go).
- [`ExtensionInfo` contains `Authors`, `Tags`, `Requires`, `Before`, `After`, and `RequiresCapabilities` slices](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/extension.go).
- [`Compatibility` contains an `Engines` map and `Platforms` slice](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/extension.go).
- [`Runtime.Use` obtains `info := ext.Info()` and later appends that value to `rt.exts`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).
- [`Runtime.Extensions` copies only the outer `[]ExtensionInfo`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/src/wago/runtime.go).

### Import metadata mutation

```go
imports := mod.Imports()
imports[0].Params[0] = wago.ValV128

again := mod.Imports()
// again[0].Params[0] now observes the caller's mutation.
```

The same applies to `Results`, `ParamTypes`, and `ResultTypes`.

### Extension metadata mutation

```go
infos := rt.Extensions()
infos[0].Tags[0] = "changed"
infos[0].Compat.Engines["wago"] = "*"

again := rt.Extensions()
// The runtime's retained inspection metadata has changed.
```

A plugin can produce the same effect by retaining and mutating the slices/map it returned from `Info()` after `Use` completes.

## Impact

- callers can corrupt subsequent inspection/CLI/registry output without any mutating API;
- concurrent inspection and mutation creates data races or concurrent map read/write failures;
- plugin ordering/provenance/capability metadata can drift after validation and registration;
- compatibility information surfaced to build/publish tooling may no longer match what `Use` originally checked;
- module signature diagnostics can be modified independently of the compiled module;
- the API violates the normal expectation that returned metadata values are caller-owned snapshots.

This is especially risky for extension metadata because validation uses one logical value while later inspection may expose a mutated value, obscuring provenance and policy decisions.

## Suggested fix

Create centralized recursive clone helpers and use them at both trust/publication boundaries.

### Extension metadata

- deep-clone `ExtensionInfo` immediately after `ext.Info()` returns, before compatibility/authorization/ordering checks;
- clone `Authors`, `Tags`, `Requires`, `Before`, `After`, and `RequiresCapabilities`;
- clone `Compat.Platforms` and every entry in `Compat.Engines`;
- store only the cloned immutable value in runtime/plugin-plan state;
- deep-clone again when returning `Runtime.Extensions()`, unless the API moves to immutable accessor types.

### Module import metadata

- clone every `ImportSpec` nested slice in `Module.Imports()`;
- keep `ValueTypeDescriptor` elements copied by value;
- use one helper shared by tests and any future serialization/inspection path.

Audit other public metadata/result structs for nested mutable containers. `DefinedTypeDescriptor`, `FunctionMetadata`, and other structural summaries already require recursive cloning whenever retained internal slices are exposed; enforce this with one package-level ownership convention rather than ad hoc outer-slice copies.

For hot inspection paths, immutable internal storage plus explicit `Clone` methods can make the allocation contract visible. Do not return internal maps/slices merely to save allocations unless the method is named/documented as a borrowed view with a strict lifetime—which is inappropriate for these general public APIs.

## Regression tests

- mutate every nested field returned by `Module.Imports`; a second call remains unchanged;
- mutate every nested field returned by `Runtime.Extensions`; a second call remains unchanged;
- have a test extension mutate its original `Info()` slices/map after `Use`; runtime state remains unchanged;
- run concurrent inspection while mutating caller-owned returned data under `go test -race`;
- cover nil versus empty slices/maps and preserve deterministic JSON output;
- add a reflection-based guard that identifies new slice/map/pointer fields in inspection structs without corresponding clone coverage.

## Acceptance criteria

- no caller mutation of returned inspection data changes runtime/module state;
- extension-provided metadata is snapshotted at registration and cannot change afterward;
- concurrent callers can freely mutate their own returned values without a race inside Wago;
- all nested mutable fields have explicit clone coverage;
- the ownership/snapshot contract is documented for public metadata APIs.

## Audit context

JairusSW:
Supersedes #424 with an atomic replacement built from current `main` and fixes every finding from its code review.

## What changed

- reject `Compile` after shutdown and snapshot hooks, instructions, configuration, imports, and capability metadata at one admission boundary;
- wait for admitted compile/module operations before stopping plugins and internal services;
- isolate shutdown callback panics behind a bounded `ErrCallbackPanic` without retaining panic values or stacks, while continuing cleanup;
- reject foreign `Module` values before policy, imports, or hooks and retain explicit `Runtime.Module` rebinding;
- bind decoded cache artifacts exactly once, close rejected native mappings deterministically, and propagate binding failures;
- remove scheduling-only function-worker policy from cache identity, bump the cache format, and reconcile cache documentation;
- deep-copy nested extension/import inspection metadata at registration, planning, storage, and return boundaries.

## Regression coverage

- deterministic compile/close lifecycle ordering and concurrent admission races;
- shutdown panics in plugin, internal, and runtime hook callbacks with retained managed-instance and funcref-token cleanup;
- capability mismatch, identical imports, closed source/destination runtimes, concurrent `Use`, and explicit rebinding;
- cached binding rejection, closed runtime, corrupt artifact repair, and decoded artifact ownership cleanup;
- worker policies `0`, `1`, `2`, `4`, and above `GOMAXPROCS` across a small corpus, byte/native-metadata equality, and subprocess warm reuse;
- concurrent caller mutation of inspection snapshots under the race detector plus recursive mutable-field guards.

## Validation

- `go test -race ./src/wago ./cli/runtime/internal/artifactcache -count=1`
- `go test ./... -count=1`
- `make lint` (`staticcheck` unavailable locally and skipped by the repository target)

Closes #388.
Closes #390.
Closes #393.
Closes #400.
Closes #401.
Closes #416.

### DIFF
--- a/cli/runtime/internal/artifactcache/cache.go
+++ b/cli/runtime/internal/artifactcache/cache.go
@@ -5,6 +5,7 @@ import (
 	"crypto/sha256"
 	"encoding/binary"
 	"encoding/hex"
+	"errors"
 	"fmt"
 	"io"
 	"os"
@@ -39,6 +40,8 @@ var defaultIdentity = sync.OnceValues(func() ([sha256.Size]byte, bool) {
 	return buildIdentity(info)
 })
 
+var closeDecodedArtifact = func(compiled *wago.Compiled) error { return compiled.Close() }
+
 // LoadOrCompile loads a matching artifact or compiles source and saves the
 // result. Cache read/write failures never prevent execution; compilation and
 // artifact validation errors retain their normal behavior.
@@ -50,6 +53,11 @@ func (cache Cache) LoadOrCompile(source []byte, config *wago.RuntimeConfig, rt *
 			if compiled.UnmarshalBinary(blob) == nil {
 				if module, err := rt.Module(compiled); err == nil {
 					return module, nil
+				} else {
+					if closeErr := closeDecodedArtifact(compiled); closeErr != nil {
+						return nil, errors.Join(err, fmt.Errorf("release rejected cached artifact: %w", closeErr))
+					}
+					return nil, err
 				}
 			}
 		}
--- a/cli/runtime/internal/artifactcache/cache.go
+++ b/cli/runtime/internal/artifactcache/cache.go
@@ -30,7 +30,7 @@ type Cache struct {
 	ReportError func(error)
 }
 
-const cacheKeyFormat = 1
+const cacheKeyFormat = 2
 
 var defaultIdentity = sync.OnceValues(func() ([sha256.Size]byte, bool) {
 	info, ok := debug.ReadBuildInfo()
@@ -108,7 +108,9 @@ func (cache Cache) path(source []byte, config *wago.RuntimeConfig) (string, bool
 		encoded[len(encoded)-1] = 1
 	}
 	encoded = binary.LittleEndian.AppendUint32(encoded, config.MemoryLimitPages())
-	encoded = binary.LittleEndian.AppendUint64(encoded, uint64(config.FunctionWorkers()))
+	// FunctionWorkers is deliberately excluded: it controls only transient
+	// compilation scheduling, and serialized output is deterministic across
+	// worker policies. The key contains only semantic/codegen dimensions.
 
 	knobs := config.OptimizationInfos()
 	encoded = binary.LittleEndian.AppendUint32(encoded, uint32(len(knobs)))
--- a/cli/runtime/internal/artifactcache/cache_binding_test.go
+++ b/cli/runtime/internal/artifactcache/cache_binding_test.go
@@ -1,0 +1,99 @@
+package artifactcache
+
+import (
+	"errors"
+	"strings"
+	"testing"
+
+	"github.com/wago-org/wago"
+)
+
+type rejectingCompileExtension struct {
+	err   error
+	calls *int
+}
+
+func (e rejectingCompileExtension) Info() wago.ExtensionInfo {
+	return wago.ExtensionInfo{ID: "test.reject-cache-bind"}
+}
+
+func (e rejectingCompileExtension) Register(reg *wago.Registry) error {
+	hooks, err := reg.ModuleCompiler()
+	if err != nil {
+		return err
+	}
+	hooks.After(func(*wago.CompileContext, *wago.Module) error {
+		(*e.calls)++
+		return e.err
+	})
+	return nil
+}
+
+func TestLoadOrCompilePropagatesCachedModuleBindingFailureOnce(t *testing.T) {
+	source := constantModule()
+	config := wago.NewRuntimeConfig().WithBoundsChecks(wago.BoundsChecksExplicit)
+	cache := Cache{Dir: t.TempDir(), Identity: []byte("runtime-a")}
+	producer := wago.NewRuntime(wago.WithRuntimeConfig(config))
+	module, err := cache.LoadOrCompile(source, config, producer)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := module.Compiled().Close(); err != nil {
+		t.Fatal(err)
+	}
+	if err := producer.Close(); err != nil {
+		t.Fatal(err)
+	}
+
+	rejected := errors.New("cached artifact rejected")
+	var calls int
+	consumer := wago.NewRuntime(wago.WithRuntimeConfig(config))
+	if err := consumer.Use(rejectingCompileExtension{err: rejected, calls: &calls}, wago.WithPluginGrants(wago.PluginCompileHooks)); err != nil {
+		t.Fatal(err)
+	}
+	defer consumer.Close()
+	oldClose := closeDecodedArtifact
+	var closedArtifacts int
+	closeDecodedArtifact = func(compiled *wago.Compiled) error {
+		closedArtifacts++
+		if err := compiled.Close(); err != nil {
+			return err
+		}
+		if _, err := wago.Instantiate(compiled); err == nil {
+			t.Error("rejected decoded artifact remained instantiable after cleanup")
+		}
+		return nil
+	}
+	t.Cleanup(func() { closeDecodedArtifact = oldClose })
+	if _, err := cache.LoadOrCompile(source, config, consumer); !errors.Is(err, rejected) {
+		t.Fatalf("LoadOrCompile error = %v, want cached binding rejection", err)
+	}
+	if calls != 1 {
+		t.Fatalf("AfterCompile called %d times, want exactly once", calls)
+	}
+	if closedArtifacts != 1 {
+		t.Fatalf("rejected decoded artifacts closed %d times, want exactly once", closedArtifacts)
+	}
+}
+
+func TestLoadOrCompileRejectsCachedArtifactOnClosedRuntime(t *testing.T) {
+	source := constantModule()
+	config := wago.NewRuntimeConfig().WithBoundsChecks(wago.BoundsChecksExplicit)
+	cache := Cache{Dir: t.TempDir(), Identity: []byte("runtime-a")}
+	producer := wago.NewRuntime(wago.WithRuntimeConfig(config))
+	module, err := cache.LoadOrCompile(source, config, producer)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer module.Compiled().Close()
+	if err := producer.Close(); err != nil {
+		t.Fatal(err)
+	}
+	closed := wago.NewRuntime(wago.WithRuntimeConfig(config))
+	if err := closed.Close(); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := cache.LoadOrCompile(source, config, closed); err == nil || !strings.Contains(err.Error(), "closed runtime") {
+		t.Fatalf("cached load into closed runtime = %v, want closed-runtime error", err)
+	}
+}
--- a/cli/runtime/internal/artifactcache/cache_test.go
+++ b/cli/runtime/internal/artifactcache/cache_test.go
@@ -112,8 +112,8 @@ func TestCacheKeyIncludesRuntimeAndCompilerConfiguration(t *testing.T) {
 	if basePath == optimizationPath {
 		t.Fatal("optimization selection did not change artifact key")
 	}
-	if basePath == workersPath {
-		t.Fatal("function-worker policy did not change artifact key")
+	if basePath != workersPath {
+		t.Fatal("scheduling-only function-worker policy changed artifact key")
 	}
 	if basePath == boundsPath {
 		t.Fatal("bounds-check mode did not change artifact key")
--- a/cli/runtime/internal/artifactcache/cache_workers_test.go
+++ b/cli/runtime/internal/artifactcache/cache_workers_test.go
@@ -1,0 +1,137 @@
+package artifactcache
+
+import (
+	"bytes"
+	"os"
+	"os/exec"
+	"runtime"
+	"slices"
+	"testing"
+
+	"github.com/wago-org/wago"
+)
+
+type countingCompileExtension struct {
+	calls *int
+}
+
+func (e countingCompileExtension) Info() wago.ExtensionInfo {
+	return wago.ExtensionInfo{ID: "test.count-cache-compile"}
+}
+
+func (e countingCompileExtension) Register(reg *wago.Registry) error {
+	hooks, err := reg.ModuleCompiler()
+	if err != nil {
+		return err
+	}
+	hooks.Before(func(*wago.CompileContext, []byte) ([]byte, error) {
+		(*e.calls)++
+		return nil, nil
+	})
+	return nil
+}
+
+func TestArtifactsAreDeterministicAcrossWorkerPolicies(t *testing.T) {
+	policies := []int{0, 1, 2, 4, runtime.GOMAXPROCS(0) + 1}
+	corpus := [][]byte{wagoEmptyModule(), constantModule(), memoryModule()}
+	for sourceIndex, source := range corpus {
+		var baseline []byte
+		var baselineEntries []int
+		for _, workers := range policies {
+			config := wago.NewRuntimeConfig().WithBoundsChecks(wago.BoundsChecksExplicit).WithFunctionWorkers(workers)
+			compiled, err := wago.Compile(config, source)
+			if err != nil {
+				t.Fatalf("source %d workers %d: %v", sourceIndex, workers, err)
+			}
+			artifact, err := compiled.MarshalBinary()
+			if err != nil {
+				compiled.Close()
+				t.Fatalf("source %d workers %d marshal: %v", sourceIndex, workers, err)
+			}
+			entries := append([]int(nil), compiled.Entry...)
+			if err := compiled.Close(); err != nil {
+				t.Fatal(err)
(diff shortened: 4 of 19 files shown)


jtenner:
Superseded by #426, which consolidates this runtime/cache work with rolling-release fixes and includes the follow-up review hardening in dda066b2, f28fdce3, and 66be6823. Closing this draft to keep #426 authoritative.

PR #429, opened 2026-08-12
railshot: improve single-pass execution and compile footprint

JairusSW:
## What changed

This keeps Railshot single-pass while improving generated ARM64/AMD64 code and reducing transient compiler memory.

- add bounded ARM64 SIMD, conversion-source, and float-constant peepholes
- remove redundant AMD64 memory32 extensions and low-32 mask work
- reuse or right-size validator, hint, control-stack, operand-stack, type, and ownership scratch
- skip irrelevant constant-preload scans
- allocate optional telemetry only when requested
- preserve strict validation and generated-artifact behavior

## Why

Wago should keep JIT compilation extremely fast and memory-predictable while making generated code competitive with wazero. These changes target local instruction selection and bounded one-pass metadata only; they do not add an IR, whole-program optimizer, unbounded cache, or execution allocation.

## Measured impact

The equivalent pre-rebase series (`39ab0f57`) was measured on native AMD64 against wazero v1.9.0 using pinned CPU7, `GOMAXPROCS=1`, fixed engine-interleaved ABBA order, 15 samples, and all 36 compile-capable corpus modules:

- direct compile: 3.335x geometric-mean speedup
- direct compile memory: 10.63x lower B/op and 4.86x fewer allocations
- full compile: 2.019x geometric-mean speedup
- full compile memory: 8.24x lower B/op and 2.95x fewer allocations
- AMD64 execution: 1.607x explicit-bounds and 1.188x guard-page geometric-mean speedup versus wazero
- Wago execution remained 0 B/op and 0 allocs/op

Representative retained ARM64 deltas include 24.6% faster UTF conversion, 21.5% faster UTF validation, and 8.1% faster nbody. The series was rebased onto current `main`; `git range-diff` reports all 30 patches unchanged.

Deterministic footprint examples:

- Ruby parallel-validation error scratch at 8 workers: 279,232 B to 192 B
- module facts: up to three fewer transient allocations
- structural type canonicalization: two fewer allocations and four bytes saved per flattened type
- dense global-hint scratch: one fewer allocation on first growth

## Validation

- `go test ./...`
- `WAGO_CORPUS_TIMEOUT=20s make test-corpus`
- `go generate ./cli/internal/settings` with a clean worktree
- explicit and guard-page corpus suites
- generated artifacts checked byte-identically for compile-memory-only candidates
- fixed ABBA benchmark campaigns; timed-out or interrupted runs were excluded rather than partially analyzed

## Scope notes

The separately preserved ARM64 memory32 UXTW and float-load-sink experiments are not included because the local ARM host was too loaded for trustworthy native timing. Candidates with unfavorable latency-to-memory tradeoffs were also excluded.


src/core/compiler/optimization/catalog.go
```diff
@@ -125,14 +162,56 @@ func (b *Bindings) Set(name string, on bool) bool {
...
+// still names the current process defaults, only deltas are installed. The
+// revision comparison, delta installation, compilation lease, and restoration
+// all share the same lock, so a concurrent Set cannot invalidate the fast path.
+// If the process defaults changed since capture, overrides is applied in full.
+func (b *Bindings) ApplySnapshot(overrides map[string]bool, snapshot Snapshot, deltas map[string]bool) (func(), error) {
 	b.mu.Lock()
+	if overrides == nil {
+		return b.mu.Unlock, nil
+	}
+	if snapshot.bindings == b && snapshot.revision == b.revision {
```
JairusSW: **[P2] Don’t let the snapshot fast path silently ignore the complete selection.**

`CompileOptions.Optimizations` is documented as the complete selection, but when the revision matches this branch never reads `overrides`; it validates and applies only `deltas`. Since `OptKnobSnapshot` and these option fields are exported, a caller can change the complete map (or add an unknown name), pass the still-matching snapshot with a missing/stale delta map, and silently compile with the process defaults instead. That is particularly risky for optimization kill switches.

Could this use an opaque coherent snapshot/selection object, or fall back to validating and applying `overrides` unless coherence can be established? Please also add cases for a changed override omitted from `deltas`, conflicting delta/full-map values, and an unknown name while the revision matches.

JairusSW (commented):
I found one functional issue in the optimization snapshot fast path, plus two non-blocking test/comment concerns. Inline comments attached.

src/core/compiler/backend/railshot/amd64/allocation_test.go
```diff
@@ -3,11 +3,49 @@
...
+				if err != nil {
+					t.Fatalf("CompileModuleWith: %v", err)
+				}
+				if !bytes.Equal(got.Code, baseline.Code) {
+					t.Fatal("native code changed across identical serial compiles")
+				}
+				benchCompiledSink = got
+			})
+			t.Logf("CompileModuleWith allocations = %.0f", allocs)
+			if allocs != tc.wantAllocs {
```
JairusSW: **[P3] Avoid exact allocation totals for whole compiler invocations.**

These counts can change across Go toolchains or due to unrelated escape/inlining changes, and this test would even fail on an improvement such as `173 -> 172`. The existing allocation tests below intentionally use conservative ceilings for that reason. Please make these upper bounds with a small margin while keeping the native-code identity assertion exact.

The same concern applies to `TestTypeDescriptorCorpusAllocations` requiring `allocs == 2`; `allocs <= 2`, together with the capacity and mutation-isolation tests, still enforces the intended coalescing without pinning compiler implementation details.

src/core/compiler/backend/railshot/amd64/emit.go
```diff
@@ -1056,7 +1063,17 @@ func (f *fn) condenseInto(e *elem, dest Reg) {
 func (f *fn) applyALU(enc aluEnc, dest Reg, right *elem, w bool) {
 	switch right.st.kind {
 	case stConst:
-		if fitsImm32(right.st.cval) {
+		// `i64.and x, 0xffffffff` is exactly a zero-extension of x's low
+		// 32 bits. `and r32, -1` performs that operation directly on x86-64 and
+		// preserves the ordinary AND's flag writes. The 64-bit form cannot encode
```
JairusSW: **Nit:** the 32-bit form does not preserve *all* flags of `and r64, 0xffffffff`. When bit 31 is set, `and r32, -1` sets `SF`, whereas the 64-bit masked result always has bit 63 clear and therefore `SF = 0`. The value result and the flags relevant to current zero tests appear fine; could this comment be narrowed so it does not claim full flag equivalence?

PR #451, opened 2026-08-14
plugin: key binary cache on resolved inputs

### LINKED ISSUE #406: [P0][Plugin Build] Include the complete build input graph in the binary cache key
## Problem

The custom plugin-runtime binary cache key does not cover all inputs passed to `go build`.

`Hash` currently includes runtime version, profile, Go version, GOOS/GOARCH, a fingerprint of the Wago source checkout, and sorted plugin module paths. It omits at least:

- `Config.BuildTag`, although the tag is passed to `go build -tags`;
- selected module versions and sums from the generated module graph;
- `go.mod`/`go.sum` contents after `go mod tidy`;
- source content of local filesystem replacements other than the Wago checkout;
- relevant build environment/toolchain dimensions that can change selected files or output.

A matching `.hash` can therefore cause `EnsureBinary` to return a stale binary built with a different plugin version, feature profile/tag, or local replacement source.

## Evidence

- [`Hash` omits `Config.BuildTag` and hashes dependency names only](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/plugin/build/module.go).
- [`ensureBinary` calculates `want := Hash(...)` before `go mod tidy`, then returns the existing binary when that hash matches and `syncBuildModule` made no new edit](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/plugin/build/module.go).
- [The actual build conditionally appends `-tags config.BuildTag`](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/plugin/build/module.go).
- [`syncBuildModule` mirrors filesystem `replace` directives, including sibling/local plugins](https://github.com/wago-org/wago/blob/ca3266ca7adca444f17d2f1f225d71854bba8ad7/cli/manager/internal/plugin/build/module.go).
- only the Wago source directory is passed through `localSourceFingerprint`; mirrored replacement targets are not.

### Build tag/profile

Build once with one tag set, then request a configuration whose `BuildTag` selects a different runtime feature surface. If runtime version/profile strings are otherwise unchanged, the binary hash can match and the old executable is reused without invoking `go build`.

### Plugin update

Update `example/plugin` from one selected version to another while the dependency path remains `example/plugin`. If `go.mod` was changed by a preceding command and `syncBuildModule` itself makes no further edit, the cached hash still contains only `example/plugin` and can reuse the old binary.

### Local replace

Edit a plugin resolved through `replace example/plugin => ../plugin`. The Wago checkout fingerprint is unchanged, the dependency path is unchanged, and the cached executable can remain stale.

## Impact

- runtime behavior does not match the lockfile/generated module or requested profile.
- security/bug fixes in an updated plugin may not reach the executable.
- local development appears nondeterministic because edits do not trigger rebuilds.
- artifact-cache identity embedded into the reused binary can describe the wrong build inputs.
- users may need `--force` or manual cache deletion without knowing why.

## Suggested fix

Compute the binary identity after module synchronization/resolution and include a deterministic build manifest.

Recommended inputs:

- runtime version, profile, exact `BuildTag`, Go/TinyGo toolchain identity, GOOS/GOARCH, CGO/build-mode values, and relevant `GOAMD64`/`GOARM64` feature level;
- generated `main.go` semantic content or its normalized dependency list;
- normalized post-`tidy` `go.mod` and `go.sum`;
- output of `go list -m -json all` or an equivalent canonical selected module graph with versions, sums, and replacements;
- a content fingerprint for every filesystem replacement that can contribute code, not only Wago;
- build flags that affect file selection/code generation.

For local replacement trees, prefer a reusable bounded source-tree fingerprint that covers tracked modifications, untracked source files, generated files used by the build, and non-Git directories. Fail closed to rebuilding when a reliable fingerprint cannot be produced.

Write the final manifest/hash only after the staged executable has built and installed successfully, using atomic replacement.

## Regression tests

- same plugin path, different selected version/sum => rebuild.
- same configuration except `BuildTag` => rebuild and expected feature surface differs.
- edit tracked and untracked files in a local plugin replacement => rebuild.
- edit a mirrored sibling replacement => rebuild.
- unchanged resolved graph/source/config => cache hit.
- failed `tidy`/build/install does not publish a new hash.
- hash is deterministic across dependency ordering and independent of irrelevant timestamps.

## Acceptance criteria

- every input capable of changing the custom runtime binary changes the cache key or conservatively forces a rebuild.
- plugin version updates and local source edits cannot return an older executable.
- build tags are represented explicitly.
- the key is computed from the resolved graph actually used by `go build`.
- tests exercise published modules, local replacements, dirty trees, and profile/tag changes.

## Audit context

JairusSW:
Closes #406.

## What changed

- Build the executable cache key after `go mod tidy` from the resolved module graph.
- Include the build tag, Go toolchain and target settings, `go.mod`, `go.sum`, generated source, module versions and sums, replacements, and every selected source or embedded file.
- Bound source fingerprint work. Inputs that cannot be proved safe are rebuilt instead of cached.
- Invalidate the old key before a build and check local inputs again before publishing the binary.
- Write generated `main.go` with atomic replacement.

## Why

The old key could describe more than one executable. A build-tag change, module update, or local source edit could therefore return an older plugin runtime. This could keep a security or correctness fix out of the running binary.

## Validation

- `GOTOOLCHAIN=go1.22.12 go test -count=1 ./cli/...`
- `GOTOOLCHAIN=go1.22.12 go test -count=1 -tags wago_runtime ./cli/...`
- `go test -race ./cli/manager/internal/plugin/build ./cli/manager/internal/plugin`
- focused `staticcheck` 2024.1.1 with Go 1.22
- `make lint` except its local staticcheck install, which is not compatible with the host Go 1.26 toolchain

The full local suite reached the staged Core 3 tests, then stopped because the pinned Release 3 interpreter and WABT tools are not installed locally. CI installs those tools.

Docs unchanged: this is an internal correctness fix with no developer workflow change.


JairusSW:
@codex review

JairusSW:
@codex review

JairusSW:
@codex review

JairusSW:
@codex review

JairusSW:
@codex review

JairusSW:
@codex review

JairusSW:
@codex review

PR #466, opened 2026-08-15
integration: bounded Railshot performance campaign

JairusSW:
## Status

**Integration campaign only — do not merge this aggregate PR.**

This draft is stacked on `jairus/more-efficient-codegen-output` so the complete campaign can be qualified together. At 148 commits and 200 changed files, it remains too broad to be a responsible merge unit. Landing requires smaller, independently reviewable PRs; keeping this draft is the correction for the review's causality and rollback blocker, not a claim that the aggregate is reviewable.

## Scope

- bounded semantic facts, rematerialization, regional residency, call effects, ABI work, generated/local machine rules, and targeted AMD64/ARM64 campaigns
- immutable per-compilation optimization policy and reusable worker scratch
- the plan is recorded in `docs/unified-railshot-performance-plan.md`
- the completed code-size finalizer remains separate from semantic optimization

## Review validation and repairs

The source-level review was checked finding by finding.

- **Optimization inventory:** schema generation, manifest validation, architecture binding validation, discovery, and docs now use the canonical optimization catalog. An equality test prevents schema/validator drift.
- **`LeafHostFunc`:** removed rather than documenting an unenforceable non-reentrancy promise. This also removes the module-wide leaf-dispatch cliff and undocumented leaf panic/error protocol.
- **Default policy:** new semantic/ABI/GC experiments are default-off behind explicit experimental controls. `entry-init-elide` remains on because it is already a current-`main` baseline optimization, not newly enabled by this campaign; it retains an explicit rollback control.
- **Result ABI:** added exhaustive Cartesian planner coverage and focused mixed-bank, dirty-high-bit, NaN, negative-zero, trap, stale-trap, fallback, reference, and v128 boundary tests.
- **Collector ABI:** compiled artifacts already carry `NativeABIVersion`, and incompatible layouts are rejected. Native GC transforms and resolved-object reuse remain experimental/default-off; unknown operations conservatively invalidate cached resolutions.
- **Prepared entry proof:** general, private direct, and isolated direct are now documented separately. Tables and GC exclude isolated entry but may remain eligible for serialized private direct execution; compiler body restrictions are stated per architecture. Linux/ARM64 ABI tests pin explicit bounds instead of assuming the Darwin default.
- **Stack fence:** both backends now validate the final frame size through a bounded retry rather than trusting body-byte count alone.
- **Machine-rule generator:** normalized identifier collisions and invalid names are rejected, generated Go is type-checked, and hostile names are tested.
- **Hint storage:** merge metadata has explicit fixed ownership instead of hiding state in spare slice capacity.
- **Exact-head correctness defects found during qualification:** repaired Go-assembly result declarations, TinyGo prepared-entry surfaces, stale trap-cell consumption, AMD64 finalizer scratch ownership, EH frame elision, Linux/ARM64 test qualification, AMD64 indirect SIMD wrapper-tail selection, the register-caller wrapper-tail result-pointer contract, and spill-slot/GC-local provenance aliasing.

The reviewed result-pointer and raw-object provenance defects are repaired at the current head. Native GC transforms and resolved-object reuse remain default-off and must still land separately with focused qualification.

## Paired deltas only

Lower is faster. These are paired medians; no standalone timings are presented.

Environment: Apple M4 Max, Darwin/ARM64, Go 1.26.5, guard-page build, `GOMAXPROCS=1`, five 500 ms samples per row, 36 matched corpus rows.

Exact revisions:

- current `main`: `bb0eeea54f00ba5396ef780058fd6d060ac0a4f9`
- measured ARM64 candidate: `cf8508d231d902a3eb9436c705962a1525435488`
- current PR head: `872770b32afb8bd66e82bf124e0894cec30eba0d`

The table remains historical aggregate evidence from `cf8508d2`, not exact-head evidence for the current default policy. Production correctness changes have landed on both backends since capture.

| ARM64 full-corpus comparison | Geometric mean | Workload-weighted | Faster rows |
| --- | ---: | ---: | ---: |
| Execution vs current main | **+20.82% slower** | **+2.18% slower** | 11/36 |
| Execution vs wazero | **+4.76% slower** | **-18.85% faster** | 20/36 |
| End-to-end compile vs current main | **+15.49% slower** | **-0.54% faster** | 6/36 |
| End-to-end compile vs wazero | **-59.28% faster (2.46x)** | **-79.45% faster (4.87x)** | 36/36 |

Existing Wago execution allocation behavior is unchanged: 36/36 rows remain zero-allocation.

This default-policy result fails the plan's whole-corpus regression gate and does **not** support an “across all corpora faster than wazero” claim. Tiny call-boundary rows dominate the geometric regression after the reviewed ABI experiments were made default-off. The weighted result is positive only because longer-running workloads improve.

Exact-head AMD64 performance deltas are still pending because `hub@hub` is timing out on SSH port 22. Older AMD64 measurements are intentionally omitted.

## Exact-head qualification

Current head: `872770b32afb8bd66e82bf124e0894cec30eba0d`

Dispatched exact-head workflows:\n\n- CI: https://github.com/wago-org/wago/actions/runs/31922403293\n- regression stress: https://github.com/wago-org/wago/actions/runs/31922404364\n\nLocal qualification at this head:

- ARM64 `go test ./...` — passed.
- `bench/go test ./...` — passed.
- AMD64 Go 1.22.12 `go test ./...` — passed after initializing the pinned `tests/spec-v3` fixture.
- AMD64 focused mixed-result indirect-tail matrix under `-race` — passed for Balanced, Size, and Embedded, including repeated nested-register-caller execution.
- AMD64 and ARM64 GC provenance invariant tests — passed; `storage` remains 64 bytes.

Both exact-head workflows completed successfully. CI attempt 1 stopped in the checkout path filter with a transient `shallow file has changed` error before tests; attempt 2 passed the complete matrix, including the focused Linux/AMD64 race lane. The aggregate stays non-mergeable regardless: exact-head paired deltas are absent, the ARM64 aggregate evidence fails the corpus gate, and production changes must be split into independently reviewable children.


JairusSW:
Performance follow-up split from this integration branch:

- #470 — ARM64: admit memory-independent prepared leaves

The child is rebased directly onto current `main` and is intentionally limited to one commit and two files. Exact Darwin/ARM64 deltas versus main are -31.24% (`mulhi`), -30.66% (`pack`), and -30.70% (`parse4`), all at 0 B/op and 0 allocs/op. The full 36-row corpus improves 2.90% by geometric mean and 0.19% workload-weighted.

Qualification also found and rejected a broader admission: a memory-free caller can still reach a memory-touching callee. The final child therefore requires call-free as well as memory-independent, with an executing transitive-memory near miss. Native bytes are unchanged; summed compile B/op changes by +0.0001%; peak RSS was 162.4 MB on main and 161.6 MB on the child.

The later shared mutex-elision experiment remains integration-only because it helped this aggregate stack but did not clear the incremental performance bar against the clean landing parent.

jtenner (commented):
Reviewed as integration evidence, not as a merge candidate. Two items need correction on the current head.

1. **AMD64 `applyMul` still releases a borrowed pinned-local address register.** In the `stMemRef` branch, both paths call `releaseMemRef(right.st)`, which correctly preserves `right.st.reg` when `memBorrow() >= 0`; the branch then unconditionally calls `f.release(right.st.reg)`. That marks a borrowed pinned-local register free and permits allocator reuse while the local is still live. This is the multiplication analogue of the comparison ownership defect fixed in #477. Remove the unconditional release/use `releaseMemRef` as the sole address cleanup, and add a minimized multiplication variant of the borrowed-local regression.

2. **The exact-head qualification in the PR body is stale.** The body identifies `872770b32afb8bd66e82bf124e0894cec30eba0d` as the current/qualified head, while the actual head is `70d04ce9f99c16e9d3e56034546b5e56df035b4e`, four commits later. I also found no PR workflow or commit-status results attached to the actual head. Refresh the exact-head section before using this aggregate as qualification evidence.

The PR already correctly states that it must not merge as an aggregate; this review does not request a merge-state change.

PR #468, opened 2026-08-15
threads: bound host-global serialization test

### LINKED ISSUE #462: [P1][Threads] Bound intermittent Darwin host-global serialization hang
## Problem

`TestThreadsHostGlobalAccessSerializesWithInvoke` intermittently hangs until the package-level 11-minute test timeout on native Darwin runners.

Observed in CI run `31893650721` on August 15, 2026, simultaneously in:

- Darwin/amd64 job `95033551760` during the guard-page suite;
- Darwin/arm64 job `95033551790` during the ordinary build-and-test suite.

Both jobs ended with:

```text
*** Test killed with quit: ran too long (11m0s).
FAIL github.com/wago-org/wago/src/wago 660s
```

The amd64 dump showed goroutines in `TestThreadsHostGlobalAccessSerializesWithInvoke` around `src/wago/threads_atomic_test.go:345` and `:366`. Earlier CI runs of the same PR passed on both Darwin architectures. The triggering commit changed only the transform-only frontend tail-result contract helper, its tests, and documentation, so this appears unrelated to that PR.

## Scope

- reproduce the host-global/invoke serialization hang under repeated Darwin execution;
- audit lock ordering and native-entry coordination in the test and runtime path;
- add bounded diagnostics or an internal deadline so a failure identifies the blocked phase rather than consuming the package timeout;
- verify the test repeatedly on both Darwin/amd64 and Darwin/arm64.

## Acceptance criteria

- repeated stress runs do not hang on either Darwin architecture;
- any genuine serialization failure terminates with a focused diagnostic well below the package timeout;
- Linux runtime-concurrency and race suites remain green.

jtenner:
## Summary

- reserve two Go scheduler Ps for the test's deliberate generated-code spin rendezvous;
- bound guest entry, guest release, and host-global completion phases with focused state diagnostics;
- document the concurrency-test requirement.

## Root cause

Generated native Wasm does not yield its P to Go's asynchronous scheduler. When this test ran in a one-P process, the guest's deliberate atomic spin occupied the only P, so the Go controller could not publish the release word. The test then consumed the package-level timeout. This explains the intermittent Darwin behavior under constrained runner scheduling.

## Validation

- `GOMAXPROCS=1 go test ./src/wago -run '^TestThreadsHostGlobalAccessSerializesWithInvoke$' -count=100`
- `go test -race ./src/wago -run '^TestThreadsHostGlobalAccessSerializesWithInvoke$' -count=20`
- `go test ./...`
- `make lint`
- `make docs-check`
- Darwin amd64 and arm64 `src/wago` test binaries cross-compiled

Closes #462

### DIFF
--- a/docs/threads-atomics-plan.md
+++ b/docs/threads-atomics-plan.md
@@ -405,8 +405,13 @@ passes the full operation/width matrix, end-of-memory and effective-address
 overflow checks for every write/RMW form, contention invariants, true-overlap
 rendezvous, waiter lifecycle tests, and 250 instantiate/invoke/close cycles.
 
-Race-focused waiter and lifecycle tests pass under `go test -race`. Test
-binaries cross-compile for linux, darwin, and windows on amd64 and arm64; Windows
+Race-focused waiter and lifecycle tests pass under `go test -race`. Tests that
+hold generated native code in a deliberate spin while a Go controller publishes a
+shared-memory release must reserve at least two `GOMAXPROCS`: generated code does
+not yield its P to Go's scheduler, so a one-P test process cannot run the controller.
+Every such rendezvous must also use bounded phase waits with state diagnostics rather
+than relying on the package timeout. Test binaries cross-compile for linux, darwin,
+and windows on amd64 and arm64; Windows
 does not advertise Threads, and this local run did not provide native Linux or
 Windows execution.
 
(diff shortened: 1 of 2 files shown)


jtenner:
@codex review

Please review the scheduler-root-cause fix and bounded phase diagnostics for the threaded host-global/native-spin serialization test.

PR #474, opened 2026-08-16
tests: add semantic corpus with exact-oracle runners

JairusSW:
Adds Wago's semantic corpus: real programs compiled to wasm, each pinned to an
upstream revision and checked against an **exact, independently verifiable
oracle**. A module that merely compiles and exits does not count as a pass.

## What's here

- `tests/semanticcorpus/` — a strict, fail-closed manifest loader and an
  exact-oracle runner (fresh instance + a second instance from the same
  compiled module, both must reproduce the oracle).
- `tests/corpora/` — per-corpus pinned artifacts, provenance, copied licenses,
  and rebuild-and-verify `build.sh` scripts:
  - **CoreMark** — published per-kernel CRCs for the 2K run (strength 2/3).
  - **BLAKE3** — 105 published test vectors across hash/keyed/derive (strength 1).
  - **QOI** — exact encode + decode vs a native reference (strength 3/5).
  - **zlib inflate** — exact decode of a native-encoded stream (strength 5).
  - **zstd decompress** — exact decode of a native-encoded frame (strength 5).
  - **LZ4** — checked in with oracle + build script, but skipped via a new
    `known_issue` field pending a wago JIT fix (see below).
- `make test-semantic-corpus` gate.

## Finding: LZ4 exposes a wago JIT discrepancy

`lz4_compress_run` (upstream LZ4 block API) is correct under wasmtime at every
optimization level (compressed size `3421`), but wago's JIT traps with a false
linear-memory OOB on `-O1`/`-O2`/`-Os` builds and silently returns a wrong size
(`75`) on an `-O0` build. Reproduced on Darwin/arm64 with default explicit
bounds checks, with 2 MiB guest memory, and with the guest stack relocated away
from data — so it is not a runner layout problem. The pinned `lz4.wasm`
(SHA-256 `00edcd94…`) is a ready reproducer. I'll chase the root cause and fix
it in a separate PR.

## Run

```sh
go test ./tests/semanticcorpus
# or: make test-semantic-corpus
```

Result: 8 passing cases, 2 skipped on `known_issue`.


jtenner (commented):
I found two fail-closed gaps in the runner/manifest boundary.

1. **Pointer exports bypass `limits.timeout_ms`.** `resolvePointer` uses unbounded `Instance.Invoke`. In single-invocation mode, the input pointer is resolved before the timeout context is even created, and the output pointer is resolved later through the same unbounded helper. Vector-mode pointer exports also ignore the existing context. A malformed or miscompiled helper export can therefore hang the semantic gate indefinitely even though the manifest says execution is bounded. Create the deadline before any guest execution, pass it through a `resolvePointerContext` helper using `InvokeContext`, and add a regression with a non-terminating pointer export.

2. **The supposedly strict manifest decoder accepts trailing content.** `LoadManifest` performs one `Decode` with `DisallowUnknownFields`, but never decodes again and requires `io.EOF`. A valid manifest followed by another JSON value—or arbitrary trailing bytes after the first value—can be accepted. Decode once more and require EOF, with a focused trailing-content test.

The exact-oracle checks, artifact hashing, copied memory outcomes, and fresh/second-instance comparison otherwise look sound.

PR #475, opened 2026-08-16
arm64: skip dead forward-merge reloads

JairusSW:
## What changes

ARM64 currently reloads pinned locals while converging a forward control-flow edge, even when the local is overwritten or never read after the merge.

This PR adds a shared, fixed-capacity next-use scan and lets ARM64 keep those locals memory-only when a bounded proof shows the reload is dead.

- Scans at most 64 Wasm operations.
- Uses fixed stack storage for at most 64 candidates.
- Allocates no per-operation state and retains no CFG or IR.
- Falls back to the existing eager reload on calls, nested control, EH, atomics, bulk operations, malformed input, capacity exhaustion, or fuel exhaustion.
- Adds the immutable `merge-next-use` option.
- `WAGO_ARM64_NO_MERGE_NEXT_USE=1` restores the previous behavior.

The architecture-neutral scan lives in `railshot/shared`; ARM64 retains ownership of register selection and merge-state changes.

## Measured effect

Base: `2f1f54faa62a6bf2a4ffe0b404f1102e19b2c7a6`  
Candidate: `9528f3bdd2829b048a5ec8b74a75a570b93169d5`  
Machine: Apple M4 Max, Go 1.26.5

### Candidate with optimization enabled vs same candidate with rollback enabled

| Measurement | Delta |
| --- | ---: |
| 36-row execution geometric mean | +0.12% |
| Rows faster | 17/36 |
| `memory_tree` execution | -0.92% |
| `fib_rec` execution | +0.37% |
| `raytrace` execution | +0.15% |
| Focused merge compile time | -1.53% |
| Focused merge compile B/op | 0.00% |
| Focused merge compile allocs/op | 0 |
| Native code across the corpus | -3,524 bytes |
| Modules with native-code growth | 0 |

The corpus recorded 844 eliminated merge reloads. All 288 execution samples remained at 0 B/op and 0 allocs/op.

The execution geometric mean is neutral and remains inside the 0.5% whole-corpus regression gate. The material result is reduced merge traffic and generated code without a measurable broad execution regression.

### Candidate vs exact main

Measured with one serialized `esbuild` full-compile run per sample in main/child/child/main order:

| Measurement | Delta vs main |
| --- | ---: |
| Full compile time | +1.19% |
| Compile B/op | +0.012% |
| Compile allocs/op | +5 |
| Peak RSS | +0.15% |

These remain inside the compiler time and memory gates.

## Validation

- `go test ./...`
- `go test ./...` in `bench/`
- focused ARM64 tests under `-race`
- shared scanner tests for overwrite, read, barrier, function end, fuel, capacity, malformed input, and deterministic fallback
- `make lint` — Go vet and Node checks passed; staticcheck was unavailable and the target reported it skipped
- `git diff --check`
- final-head GitHub CI: 27 passed, 1 intentionally skipped

### DIFF
--- a/schema.json
+++ b/schema.json
@@ -91,6 +91,7 @@
               "local-slot-order",
               "loop-precheck",
               "loop-region-pins",
+              "merge-next-use",
               "multi-bounds-cert",
               "olddest-rhs-sink",
               "reg-abi",
--- a/src/core/compiler/backend/railshot/arm64/compile.go
+++ b/src/core/compiler/backend/railshot/arm64/compile.go
@@ -44,6 +44,11 @@ var uxtwAddEnabled = os.Getenv("WAGO_ARM64_NOUXTW") != "1"
 // nodes. WAGO_ARM64_NOPROVENANCE=1 retains the pre-facts path for A/B checks.
 var valueFactsEnabled = os.Getenv("WAGO_ARM64_NOPROVENANCE") != "1"
 
+// mergeNextUseEnabled avoids forward-edge local reloads when bounded lookahead
+// proves the local is overwritten or dead before its next read. Loop and EH
+// targets stay conservative. WAGO_ARM64_NO_MERGE_NEXT_USE=1 restores eager loads.
+var mergeNextUseEnabled = os.Getenv("WAGO_ARM64_NO_MERGE_NEXT_USE") != "1"
+
 // sharedTrapUnwindEnabled lets Size/Embedded functions replace repeated
 // terminal trap-unwind tails with one function-local cold tail. The hot trap
 // checks and the Speed/Balanced layouts are unchanged.
--- a/src/core/compiler/backend/railshot/arm64/control.go
+++ b/src/core/compiler/backend/railshot/arm64/control.go
@@ -1401,7 +1401,7 @@ func (f *fn) opElse() error {
 	return nil
 }
 
-func (f *fn) opEnd() error {
+func (f *fn) opEnd(r *wasm.Reader) error {
 	last := len(f.ctrl) - 1
 	fr := f.ctrl[last]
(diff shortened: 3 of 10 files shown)


jtenner (commented):
Reviewed the bounded next-use scanner, candidate construction, target-state weakening, ordinary forward-edge convergence, and the `if`-without-`else` false-edge stub path. I did not find a correctness defect.

The proof remains conservative where it needs to be: malformed immediates, fuel/candidate exhaustion, calls, nested or exceptional control, bulk/atomic operations, and uncertain `end` boundaries all retain the existing eager reload. For proven-dead candidates, changing the merge target from `lsStackReg` to `lsMem` is valid because the canonical slot is already current on the incoming edge. The focused scanner and executing merge tests cover the important read-before-write, overwrite-before-read, dead-at-return, and fallback cases.

Current final-head CI is green. No blocking review comments from me.

jtenner:
@codex review

jtenner:
@codex review

jtenner (approved):
Adversarial final-head review completed at e39f5748.

I found and fixed one correctness issue during review: the merge next-use scanner followed the fallthrough path across br_on_null/br_on_non_null, which could classify a local as dead while the taken edge still needed it. Commit e39f5748 routes the scan through the shared control-boundary classifiers and adds regressions for both reference branch opcodes.

I re-reviewed candidate construction, target weakening, ordinary block/if merges, if-without-else false-edge stubs, local-base remapping, terminal/fuel/capacity/malformed fallbacks, call/EH/GC/bulk/atomic barriers, and the fixed-storage bounds. No further issues found. Codex also reported no issues on e39f5748.

Proof includes the focused shared race test, ARM64 cross-compilation, bench suite, lint, full platform/conformance CI, and a successful rerun of the unrelated flaky amd64 race timeout.

PR #476, opened 2026-08-18
wago: expose callback-scoped guest storage to host imports

jtenner:
## Summary

Add general callback-scoped guest-storage APIs for synchronous host imports. These APIs expose the storage Wago already owns without introducing a system-interface-specific layer.

This is motivated by the Facet system-interface specification:

- https://github.com/jtenner/facet-spec
- https://github.com/jtenner/facet-spec/blob/main/SPEC.md

Facet requires explicit indexed Memory32/Memory64 access, zero-copy typed Wasm GC arrays, nested GC-array scatter/gather validation, and caller-selected GC array result types. The new Wago APIs are deliberately generic so other plugins and host libraries can use the same facilities.

### Indexed linear memory

Add `GuestStorageHostModule.WithGuestStorage` and `GuestStorage` with:

- `MemoryInfo(index)` for the exact Memory32/Memory64 address form and current byte length
- `MemoryRange(index, offset, length, access)` for checked callback-scoped zero-copy ranges
- support for arbitrary memory indexes rather than implicit memory 0

### Wasm GC array borrowing

Add callback-scoped opaque `GuestGCRef` values and:

- `GCRef` to resolve an opaque GC host-argument token
- `GCArrayInfo` for dynamic storage, length, mutability, and local type identity
- `GCArrayBytes` for bounded zero-copy numeric/v128 payload views
- `GCArrayRef` for nested collector-reference arrays without manufacturing retained public tokens

The collector exposes only a bounded logical array payload view; object headers, backing pointers, and collector layouts remain private.

### Exact import types and GC result allocation

The active host callback now carries the exact structural parameter/result descriptors already computed by Wago's host bridge. `GuestStorage` exposes them through `ImportParamType`, `ImportResultType`, and `DefinedType`.

Add `GuestGCArrayAllocatorHostModule.NewGCArrayResult` for allocating the exact caller-selected defined numeric/v128 array result type. The host initializes the private rooted payload before Wago publishes the normal opaque GC result token. This supports immutable array results without exposing a post-publication mutation path.

### Lifetime and re-entry safety

All borrowed slices and callback-scoped GC refs expire when `WithGuestStorage` returns. Nested borrows fail closed. `Instance.InvokeFromHost` is rejected while a direct guest-storage borrow is active so re-entered Wasm cannot grow memory or trigger a moving collection underneath a live host slice.

### Documentation

Add `docs/host-guest-storage.md` and link the API from the plugin and feature documentation. The docs describe lifetime, locking, indexed-memory, GC-array, exact-type, allocation, and re-entry contracts and link back to Facet as a motivating consumer.

## Validation

The branch includes focused tests for:

- callback lifetime and expired-view rejection
- nested-borrow rejection
- re-entry rejection while storage is borrowed
- exact import parameter/result descriptors
- nonzero indexed-memory access and isolation
- Memory64 metadata/range access
- collector logical array payload views
- exact immutable GC-array result allocation and read-only borrowing

The implementation was compile-checked during the staged branch migration. This PR is left draft while the normal repository CI runs the complete test matrix.


jtenner (commented):
Found one correctness/API-contract issue.

**Immutable GC arrays remain writable through `GuestStorageRead`.** `GCArrayBytes` rejects `GuestStorageWrite` when `!info.Mutable`, but a read borrow returns the same directly aliased `payload.Bytes` as a mutable `[]byte`. A host can therefore request `GuestStorageRead` for an immutable array and write through the returned slice, bypassing the Wasm array mutability invariant and the documented read-only contract. Go slices do not provide read-only enforcement.

Please change the API so immutable arrays cannot expose a directly mutable alias—for example, copy immutable reads, expose read-only accessor/callback operations rather than `[]byte`, or restrict zero-copy byte views to mutable arrays—and add a regression that attempts to mutate the value returned from a nominal read borrow and verifies that immutable guest state cannot change.

I also reviewed the callback lifetime, re-entry exclusion, indexed-memory bounds checks, collector locking, and private result-allocation path. I did not find another definite correctness issue in those paths.

jtenner:
Addressed the immutable-array aliasing issue in 58b1916c. `GCArrayBytes(..., GuestStorageRead)` now returns a detached copy for immutable numeric/v128 arrays, while mutable arrays remain zero-copy. The regression mutates the returned read slice and verifies a second read still observes unchanged guest state. The API docs now call out the one allocation and payload-sized copy for immutable reads.

PR #477, opened 2026-08-20
railshot: preserve borrowed compare memrefs

jtenner:
## Summary

- preserve pinned-local address registers borrowed by deferred comparison loads
- materialize non-foldable compare loads into owned registers on amd64 and arm64
- add minimized backend regressions for both architectures

This fixes the LZ4 compression failure found while validating the semantic corpus in #474. The bad lowering corrupted local 0 and eventually produced a false linear-memory out-of-bounds trap instead of returning `3421`.

The test and fix are combined because the reproducer directly exercises this small cross-backend ownership correction.

## Validation

- `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null go test ./...`
- `go test -tags wago_guardpage ./src/core/compiler/backend/railshot/amd64 ./src/core/runtime/... ./src/wago/...`
- `cd bench && go test ./...`
- `make lint` (`staticcheck` unavailable locally; remaining checks passed)
- Linux/arm64 backend test binary cross-compiles
- exact LZ4 fixture returns `3421`, matching wazero, under explicit and signal bounds modes

## Performance / footprint

Correctness-only ownership fix. It adds no runtime data structures or Go allocations; no benchmark was run.

## Docs

Unchanged: internal backend correctness fix with no developer workflow impact.

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/emit.go
+++ b/src/core/compiler/backend/railshot/amd64/emit.go
@@ -822,10 +822,13 @@ func (f *fn) condenseCompare(node *elem, dest Reg) Reg {
 			if memRefFoldable(right.st, w) {
 				f.a.AluIdx(cmpRMcode, L, RBX, right.st.reg, right.st.memDisp(), w)
 			} else {
-				f.loadMemRef(right.st.reg, right.st)
-				f.cmpRR(L, right.st.reg, w)
+				// A narrow/wide mismatch needs a register value. Preserve a pinned
+				// local whose register is only borrowed as the load address.
+				r := f.memRefValue(right.st)
+				f.cmpRR(L, r, w)
+				f.release(r)
 			}
-			f.release(right.st.reg)
+			f.releaseMemRef(right.st)
 		}
 	}
 	f.pinned = f.pinned.remove(L)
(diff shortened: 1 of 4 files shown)


jtenner:
@codex review

Please review the borrowed deferred-memref ownership fix, especially pinned-local register preservation, matching amd64/arm64 release behavior, and the minimized regressions.

jtenner:
Post-merge follow-up from the same ownership audit: the comparison fix itself is sound, but the analogous AMD64 multiplication path still violates the borrowed-address contract.

In `applyMul`, the `stMemRef` case calls `releaseMemRef(right.st)` and then unconditionally calls `f.release(right.st.reg)`. For `memBorrow() >= 0`, `releaseMemRef` intentionally preserves the pinned local's borrowed address register; the trailing release then marks that register free anyway, permitting allocator reuse while the local remains live. ARM64 does not have this extra release.

Please remove the unconditional address-register release/use `releaseMemRef` as the sole cleanup, and add a multiplication analogue of `TestBorrowedLocalSurvivesDeferredCompareLoad`. This defect was outside the comparison-specific diff, but remains on `main` after this merge.

PR #485, opened 2026-08-20
amd64: preserve applyMul borrowed memrefs

### LINKED ISSUE #478: [P0][Compiler] Preserve borrowed memref ownership in amd64 applyMul
## Summary

The AMD64 `applyMul` lowering can release a register that is only **borrowed** from a pinned local's deferred memory reference. This can allow the register allocator to reuse that register while the local is still live.

This was found while auditing the ownership fix merged in #477 (`railshot: preserve borrowed compare memrefs`). The comparison-specific fix is sound, but multiplication still has the analogous ownership defect on `main`.

## Problem

In the `stMemRef` path of AMD64 `applyMul`, cleanup currently performs both:

```go
f.releaseMemRef(right.st)
f.release(right.st.reg)
```

`releaseMemRef` already encodes the ownership distinction. When `right.st.memBorrow() >= 0`, the address register belongs to a pinned local and must remain live; `releaseMemRef` intentionally does **not** release it.

The unconditional trailing `f.release(right.st.reg)` defeats that ownership rule and marks the borrowed register free anyway.

That creates the same class of allocator corruption addressed for deferred comparison loads in #477: a later allocation may reuse the still-live pinned-local register and corrupt subsequent execution.

ARM64 does not appear to have the corresponding extra release in its multiplication path.

## Suggested fix

Use `releaseMemRef(right.st)` as the sole cleanup for the memory-reference address register. Do not separately release `right.st.reg` when ownership may be borrowed.

## Regression coverage

Add a minimized AMD64 multiplication analogue of the borrowed-local regression added in #477, e.g. a test where:

1. a pinned local remains live,
2. a deferred memory operand borrows that local's address register,
3. multiplication consumes the memory operand,
4. register pressure causes subsequent allocation,
5. the pinned local is read again and must retain its original value.

The test should fail with the current unconditional release and pass once ownership is preserved.

## References

- #477 — comparison-path ownership fix
- #466 — aggregate Railshot campaign review where this multiplication analogue was also noted

jtenner:
## Summary

- remove the duplicate AMD64 `applyMul` address-register release
- preserve pinned-local ownership for borrowed deferred memory operands
- cover folded/materialized and borrowed/owned cleanup paths
- add an executing explicit/guard-bounds multiplication regression

Fixes #478.

## Correctness

`releaseMemRef` is the sole owner-aware cleanup for an `stMemRef` address. It releases allocator-owned addresses and deliberately preserves registers borrowed from pinned locals. The removed trailing `f.release(right.st.reg)` bypassed that distinction.

## Adversarial review

1. Audited every AMD64 `releaseMemRef` consumer for a second unconditional address release; `applyMul` was the sole remaining duplicate.
2. Compared AMD64 multiplication cleanup against ARM64 and the corrected AMD64 ALU/compare paths.
3. Mutation-tested the regression by restoring the bad release: both borrowed folded and borrowed materialized cases fail.
4. Verified owned folded/materialized operands still release their address registers.
5. Exercised the guest multiplication shape in explicit and guard-page bounds modes and under the race detector.

## Tests

- `go test ./src/core/compiler/backend/railshot/amd64 -count=1`
- `go test -tags=wago_guardpage ./src/core/compiler/backend/railshot/amd64 -count=1`
- `go test -race ./src/core/compiler/backend/railshot/amd64 -run 'TestApplyMulMemRefOwnership|TestBorrowedLocalSurvivesDeferredMultiplyLoad' -count=1`
- `go test ./src/core/compiler/backend/railshot/amd64 -run 'TestApplyMulMemRefOwnership|TestBorrowedLocalSurvivesDeferredMultiplyLoad' -count=20`
- `GIT_CONFIG_COUNT=1 GIT_CONFIG_KEY_0=tag.gpgSign GIT_CONFIG_VALUE_0=false go test ./... -count=1`
- `make lint` (staticcheck unavailable locally; CI installs it)

The first unqualified `go test ./...` attempt inherited the host's signed-tag Git configuration and failed only registry tests that create lightweight test tags. Re-running with `tag.gpgSign=false` passed the full repository. No product code was implicated.

## Performance and footprint

The implementation removes one compile-time bookkeeping write and emits no additional native instructions or runtime data. No hot-path or footprint regression is expected.

## Docs

Unchanged: this is an internal compiler ownership correction with no developer workflow or public API impact.

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/emit.go
+++ b/src/core/compiler/backend/railshot/amd64/emit.go
@@ -1191,7 +1191,6 @@ func (f *fn) applyMul(dest Reg, right *elem, w bool) {
 			f.release(r)
 			f.releaseMemRef(right.st)
 		}
-		f.release(right.st.reg)
 	}
 }
 
--- a/src/core/compiler/backend/railshot/amd64/memref_borrow_test.go
+++ b/src/core/compiler/backend/railshot/amd64/memref_borrow_test.go
@@ -67,17 +67,35 @@ func TestBorrowedLocalSurvivesDeferredCompareLoad(t *testing.T) {
 	}
 }
 
-func TestApplyMulPreservesBorrowedMemRefOwnership(t *testing.T) {
-	owner := &elem{}
-	f := fn{a: &x86.Asm{}}
-	f.regUser[R12] = owner
-	f.pinnedLocalMask = maskOf(R12)
-	right := &elem{kind: ekValue, st: memRefStorage(R12, 0, 4, false, false, 0)}
+func TestApplyMulMemRefOwnership(t *testing.T) {
+	for _, tc := range []struct {
+		name   string
+		size   int
+		borrow int
+		want   bool
+	}{
+		{name: "borrowed/foldable", size: 4, borrow: 0, want: true},
+		{name: "borrowed/materialized", size: 1, borrow: 0, want: true},
+		{name: "owned/foldable", size: 4, borrow: -1},
+		{name: "owned/materialized", size: 1, borrow: -1},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			// regUser is a canary for release: borrowed address ownership must be
+			// untouched, while an allocator-owned address must become available.
+			owner := &elem{}
+			f := fn{a: &x86.Asm{}}
+			f.regUser[R12] = owner
+			if tc.borrow >= 0 {
+				f.pinnedLocalMask = maskOf(R12)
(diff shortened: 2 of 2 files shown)


jtenner:
@codex review

Please review the AMD64 borrowed deferred-memref ownership fix adversarially. Focus on `applyMul` folded versus materialized cleanup, preservation of borrowed pinned-local registers, release of allocator-owned addresses, regression validity, and parity with the ALU/compare and ARM64 paths.

PR #489, opened 2026-08-21
plugin: support per-caller GC imports with one HostFunc

jtenner:
## Summary

Make declarative Runtime plugin host imports GC-aware without turning them into a shared `HostFuncRef`.

A plugin still registers one generic Go `HostFunc`. For every call:

- the exact Wasm parameter/result descriptors come from the importing `Compiled` module;
- the collector and GC domain come from the calling `Instance`; and
- collector objects cross the callback boundary only as temporary opaque tokens resolved through `GuestStorage.GCRef` into callback-scoped `GuestGCRef` handles.

This allows the same plugin import to serve simultaneously live modules with structurally different caller-defined GC types and different Runtime collector domains.

### One callback type, binding-local GC authority

Every declarative plugin callback remains the public `HostFunc` type in the Runtime import registry, including functions whose signatures contain GC references. There is no private GC-only callback wrapper and no second plugin host-function representation.

The Runtime keeps authority on the exact import binding instead:

- it validates the plugin's registered slot ABI against the importing module during Runtime import resolution;
- only GC-bearing imports resolved through that Runtime's plugin registry are recorded in a per-instance set keyed by function-import index; and
- host activation and argument rooting consult that set while taking the exact signature and collector domain from the calling instance.

Copying a plugin callback into an ordinary `Imports` map copies executable Go code, but not Runtime GC authority, so direct or foreign-Runtime GC use still fails closed. Scalar plugin imports need no provenance sidecar.

`Runtime.NewHostFuncRef` and `Runtime.NewGCHostFuncRef` remain monomorphic. Their exact structural-signature, collector-domain, and funcref identity checks are unchanged; both already use the same `HostFunc` callback and `HostFuncRef` handle types.

### Exact caller types and GC domains

The synchronous dispatcher continues to use `Compiled.importFuncSigs`, `Compiled.Types`, and `exactFuncSignatureView` for the active import. `GuestStorage.ImportParamType`, `ImportResultType`, and `DefinedType` therefore expose the calling module's exact descriptors.

Boundary modules now request:

- exact native GC callsite root metadata when local Wasm functions can leave native frames parked;
- the native GC ABI metadata needed by the host boundary; and
- a Runtime-owned collector domain selected from the calling module's canonical GC type graph.

An import-only module has no local native Wasm frame. For this case Wago uses one immutable exact empty root set while it still creates the Runtime GC domain. Modules with local Wasm code still require concrete exact root maps and fail closed when those maps are unavailable.

Different incompatible caller type graphs receive different Runtime collector domains. The plugin binding is never permanently attached to the first one.

The collector-reference classifier resolves indexed heap types in the containing module and excludes funcref, externref, exnref, i31-only, and bottom-only signatures from collector-object boundary admission.

### Real GC boundary handling

A collector-reference host import is treated as a real GC boundary even when the module contains no `array.new` or `struct.new` instruction.

Both complete native backends generate exact callsite roots when local Wasm code can call across this boundary:

- Linux/amd64;
- Linux and Darwin arm64.

Callsite-only root metadata is accepted for a true collector-reference boundary with no Wasm allocation sites. Import-only modules use the exact empty root set because no local native Wasm frame can exist. Generic GC modules without the required exact roots retain fail-closed behavior.

The compiled artifact codec persists and validates the native GC ABI and callsite root maps for boundary modules. Import-only modules reconstruct the exact empty root set from validated compiled metadata rather than serializing a fictitious callsite.

### GC reference representation

Plugins never receive raw internal `gc.Ref` values or object pointers.

- Non-null object arguments become temporary opaque `uint64` tokens.
- `GuestStorage.GCRef(token)` returns a Runtime-, instance-, domain-, and view-checked `GuestGCRef`.
- Null and i31 values preserve their Wasm representation and exact type validation.
- Results accept null, valid i31 values, or tokens created through the active Wago host API.
- Raw compact object references returned by a plugin remain invalid.
- Temporary argument/result tokens and roots are released after the callback handoff.

### `NewGCArrayResult` and zero-copy I/O

`GuestGCArrayAllocatorHostModule.NewGCArrayResult` works for boundary modules. It allocates the caller-selected exact array type in the calling instance's Runtime GC domain, roots it, permits initialization before immutable publication, and returns only an ephemeral host-result token.

`GuestStorage.GCArrayBytes` remains the zero-copy path for mutable numeric arrays. It returns the real payload while `WithGuestStorage` holds native execution and collector mutation stable. No raw pointer API was added. Reference arrays still reject raw byte access, and immutable arrays still reject writable access.

Scalar-only plugin imports remain ordinary `HostFunc` bindings and do not acquire GC root maps, collectors, token bookkeeping, or a scalar-call map lookup.

## Tests

Added integration coverage for:

- boundary-only nullable GC result;
- boundary-only nullable GC parameter;
- import-only collector-reference plugin declarations with an exact empty root set and a live Runtime GC domain;
- non-null host-allocated `array<i8>` result;
- object round-trip back into a plugin through `GuestStorage.GCRef`;
- zero-copy mutable array writes visible to Wasm;
- collection while a Wasm GC local remains live across a scalar host callback;
- mutable and immutable caller-selected arrays;
- two structurally non-equivalent exact array types alive simultaneously;
- distinct Runtime collector domains using the same plugin import;
- both instance/module close orders;
- codec round-trip and missing-root rejection;
- wrong plugin ABI, struct-vs-array contract, wrong array storage, non-null/null mismatch, immutable writes, forged/stale/cross-view handles, raw compact results, copied callbacks without Runtime authority, and foreign Runtime domains;
- direct low-level GC `HostFunc` rejection;
- scalar plugin imports staying on the ordinary path;
- temporary token/root cleanup;
- existing `NewGCHostFuncRef` domain and exact-type regressions.

The integration tests run on the supported complete Core 3 backends, including native arm64 CI.

`wago-facet` is also pinned to the production source commit for the import-only fix. Its integration test performs a real Facet `args_read_array_i8` call, receives `"alpha"` as a caller-selected GC `array<i8>`, reads byte 0 in Wasm, and verifies `97` (`'a'`).

## Validation

The main implementation was validated with:

- `go test ./src/wago -count=1`
- `go test -tags wago_guardpage ./src/wago -run 'TestPluginGCHostImports|TestPluginGCHostImport|TestPluginScalarHostImportStaysOrdinary' -count=1`
- full repository `go test ./... -count=1` with test-local Git tag signing disabled for registry fixtures
- focused new and existing GC host tests with `-count=20`
- focused race tests
- full CI race command: `go test -race -count=1 ./src/wago ./src/core/runtime ./tests/runtimeconcurrency`
- `make test`
- `make test-guard`
- `make test-concurrency`
- `make docs-check`
- `make lint`
- `make spec3`
- `make spec3-signals`
- Linux/arm64 and Darwin/arm64 `go test -c` cross-builds for `./src/wago`

The import-only regression is covered by `TestPluginGCHostImportOnlyModuleUsesEmptyExactRootSet`. Wago regression stress is green on the current head, and `wago-facet` CI is green against the production fix commit.

All temporary repair workflow and trigger files have been removed from the final PR tree.


jtenner:
Expanded Facet conformance coverage exposed a native arm64 GC-root corruption in this PR's synchronous host-call path.

Reproducer: `facet-spec` head `5cf8ebae873df5edb3a81fe2eda9f4cdcd58a2fc`, test `spec/tests/filesystem/dir-iterate-rewind.wast`. The Wago-Facet native arm64 run fails with:

```
WasmGC struct helper: generic GC frame-root offset 56 contains non-compact reference 0xff...
```

amd64 passes the same 143/143 suite.

I traced offset 56 to the first live GC local in the WAST. The corruption appears to come from `railshot/arm64.callHostSync`: it calls `f.flush()`, then uses caller-saved X9/X11 as control-frame/argument scratch, and only later calls `f.spillLocalsForCall()`. The arm64 GP local pin pool can assign X9/X10/X11 even in call-making functions, so a dirty pinned local can be overwritten by host-call setup and then the scratch value is written into its canonical local slot. The exact GC root scanner subsequently sees that pointer-sized scratch value instead of a compact GC reference.

The intended invariant is already stated in `callHostSync`'s comment: caller-saved pinned locals must be homed before the transition. The minimal correctness fix is to spill pinned locals immediately after `f.flush()` and before `storePinnedGlobals` / `storeModuleGlobals` / any X9-X11 host-call setup, then remove the later spill (or leave it as a no-op). It would also be prudent to audit the void `callHost` path: its comment still says pinned locals live only in X19-X23, while the current pin pool can include X9-X11.

I did not weaken the Facet test; this is a real architecture-specific regression caught by the expanded WAST matrix.

PR #491, opened 2026-08-23
arm64: preserve pinned locals across host-call scratch

### LINKED ISSUE #490: arm64: synchronous host-call setup can clobber pinned GC locals before spill
## Summary

On arm64, `callHostSync` may overwrite a live register-pinned Wasm local before `spillLocalsForCall()` homes it to the native frame. If that local is a collector reference tracked by the exact GC root map, a later allocating GC helper can observe a non-compact pointer in the published root slot and fail closed.

This was exposed by the expanded Facet conformance suite on native arm64. The amd64 backend does not reproduce the failure.

## Failure

`filesystem/dir-iterate-rewind.wast` keeps a GC-array local live across a synchronous host callback and later reaches an allocating GC helper. The root scanner reports a frame slot containing a host-call scratch pointer instead of a compact `gc.Ref`, e.g.:

```text
WasmGC struct helper: generic GC frame-root offset 56 contains non-compact reference 0xff...
```

## Root cause

In `src/core/compiler/backend/railshot/arm64/call.go`, `callHostSync` currently:

1. flushes operands;
2. synchronizes globals;
3. reuses X11 as the control-frame pointer and X9 as argument scratch while marshaling host-call arguments;
4. only then calls `spillLocalsForCall()`.

The arm64 local-pin policy can assign call-making functions to X9/X10/X11. Therefore a dirty local in one of those registers can be clobbered by host-call setup before its value is written to the frame. The subsequent spill stores the scratch pointer as the local value.

This contradicts the function's own contract comment, which notes that the extended local-pin pool includes caller-saved X8-X11 and that pinned locals must be homed before the transition.

## Fix

Move `spillLocalsForCall()` so it runs immediately after operand flush, before any host-call scratch register reuse (`storeModuleGlobals(X9)`, loading X11 with the control-frame pointer, or argument marshaling). Remove the later spill.

Add a native arm64 regression that keeps a GC local live across a returning scalar host call and then performs another allocating GC operation, proving the exact root slot still contains a compact collector reference.

The async/void `callHost` path should also be audited because its comment still assumes pinned locals only live in X19-X23.

## Context

Discovered while expanding Facet WAST API coverage to 261/261 declared and directly invoked imports. The conformance case is intentionally not weakened; this is a Wago backend correctness issue.

jtenner:
Fixes #490.

## Summary

Preserve arm64 register-pinned locals before host-call lowering reuses caller-saved X9-X11 as fixed scratch registers.

The arm64 extended local-pin pool can assign live locals to X9, X10, and X11. Both host-call paths also use those registers for fixed host-call bookkeeping. The synchronous path previously homed pinned locals only after module-global synchronization, control-frame setup, and argument marshalling had already reused X9/X11. A live local could therefore be overwritten and the scratch value subsequently written into its canonical frame slot.

The async/void path carried the same stale assumption that pinned locals only occupied X19-X23. `flushSuffix` fully canonicalizes the operand stack first, so the async path can safely home pins before reusing X9-X11.

## Fix

- `callHostSync` now calls `spillLocalsForCall()` immediately after the operand flush, before `storeModuleGlobals(X9)`, control-frame setup, or argument marshalling.
- The old late synchronous spill is removed.
- `callHost` now homes pinned locals immediately after `flushSuffix`, before X9-X11 are used by the host-call log.
- The legacy non-STACK_REG model restores those pins eagerly after the async sequence; STACK_REG retains its existing lazy recovery behavior.

This preserves the extended pin bank rather than disabling X9-X11 pinning.

## Regression

Adds native-arm64 `TestARM64HostCallPreservesExtendedPinnedLocals`. It keeps eight hot i32 locals live across a returning host call, forcing the extended pin bank beyond X19-X23, then verifies all local values survive by checking their sum is 108.

The focused native arm64 regression passed on the repair head. The regression-stress run also passed GC hardening in arm64 explicit/guard-page and amd64 explicit/guard-page modes.

## Context

The bug was discovered by the expanded Facet conformance suite on native arm64: an exact WasmGC root scan observed a host-call scratch pointer in a live GC-local frame slot instead of a compact collector reference. amd64 did not reproduce the failure.

The final PR diff contains only the arm64 backend fix and its regression test.


jtenner:
@codex please review the latest changes, including the ARM64 async host-call pin preservation and regression coverage.

PR #500, opened 2026-08-28
wasm: bound initial vector allocations

jtenner:
Small correctness fix for malformed module decoding.

Summary:
- cap initial decoder vector capacity hints at 1,024 elements
- apply the same bound to generic and byte-backed/manual decode paths
- preserve normal growth for valid larger vectors

Footprint:
- for 2 MiB remaining input and a 128-byte decoded element, the initial request drops from 256 MiB to 128 KiB

Testing:
- `go test ./src/core/compiler/...`


jtenner:

jtenner:
@codex review

PR #502, opened 2026-08-28
wasm: enforce GC supertype ordering

jtenner:
Small correctness fix for Core GC type validation.

Summary:
- reject subtype declarations with more than one direct supertype
- reject forward and self supertype references inside a recursive group
- retain the existing finality, kind, and structural subtype checks

Testing:


jtenner:

PR #505, opened 2026-08-28
policy: charge full no-max memory32 reserve

jtenner:
Small correctness fix for memory policy accounting.

Summary:
- account for the full 65,536-page Core memory32 limit when no maximum is declared
- preserve the separate staged 65,535-page memory64 implementation ceiling
- add an exact 4 GiB reservation regression test

Testing:
- `go test ./src/wago -run TestDeclaredMemoryMaxBytesChargesFullNoMaxMemory32Reservation`

The broader `go test ./src/wago` run reached unrelated pinned-corpus tests that are unavailable in the isolated worktree.


jtenner:

PR #512, opened 2026-08-28
policy: reject unenforced invoke durations

jtenner:
Small correctness fix for policy admission.

Summary:
- reject nonzero `MaxInvokeDuration` values while the low-level call path cannot enforce them
- return `ErrPermissionDenied` instead of silently admitting a weaker policy
- document the strict current behavior on the public field

Testing:
- `go test ./src/wago -run '^(TestPolicyCapabilityAllowDeny|TestPolicyRejectsUnenforcedInvokeDuration)$'`

Docs unchanged: the public API comment records the behavior.


jtenner:

PR #514, opened 2026-08-28
wasm: reject overflowing byte ranges

jtenner:
Small correctness fix for binary reader bounds checks.

Summary:
- validate requested byte counts against the remaining range without integer addition
- prevent an overflowing position-plus-length from reaching a slice-bounds panic
- add a maximum-int regression case after a nonzero cursor

Testing:
- `go test ./src/core/compiler/wasm -run '^(TestReaderBytesRejectsOverflowingRange|TestDecodeRejectsHugeVectorLengthWithoutLargeAllocation)$'`


jtenner:

PR #515, opened 2026-08-28
wasm: share owned name-section payload

jtenner:
Small correctness and footprint fix for preserved name sections.

Summary:
- keep strict structured `name` decoding unchanged
- allocate one owned raw payload after validation succeeds
- share those bytes between `RawNameSecPayload` and the matching custom-section record

Footprint:
- removes one full-payload allocation; a 16 MiB name payload retains 16 MiB less duplicate storage

Testing:
- `go test ./src/core/compiler/wasm -run '^(TestCustomNameSectionPreserved|TestValidateByteBackedModuleStrictNameSection|TestDecodeNameSectionStrictness)$'`

Docs unchanged: this is an internal footprint fix with no workflow impact.

### DIFF
--- a/src/core/compiler/wasm/decode.go
+++ b/src/core/compiler/wasm/decode.go
@@ -66,9 +66,12 @@ func decodeSection(m *Module, r *reader, id byte) error {
 				return err
 			}
 			m.NameSec = ns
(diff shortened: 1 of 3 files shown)


jtenner:

PR #518, opened 2026-08-28
wasm: validate catch payloads without copies

jtenner:
Small correctness and footprint fix for `try_table` validation.

Summary:
- compare tag parameters directly against catch-label types
- validate the optional exception reference without building a concatenated slice
- share the same allocation-free check across AST and byte-backed paths

Footprint:
- 100 validations of a 256-parameter catch now report 0 allocations in the regression test

Testing:

Docs unchanged: this is an internal validation-footprint fix.


jtenner:

PR #520, opened 2026-08-28
wasm: index GC type groups once

jtenner:
Small correctness and validation-cost fix for GC subtype metadata.

Summary:
- build one flat subtype index and one recursive-group base table per module validator
- replace repeated scans of all preceding type groups with constant-time lookups
- reuse the same immutable index throughout validation

Benchmark:
- 4,096-group flat lookup: 2.65–3.65 ns/op, 0 B/op, 0 allocs/op on the local Ryzen 7 8845HS run

Testing:
- `go test ./src/core/compiler/wasm -run '^TestGCTypeIndexFlattensGroupsOnce$' -bench '^BenchmarkGCFlatTypeLookupManyGroups$' -benchtime=200ms -count=3`

Docs unchanged: this is an internal validation-performance fix.

### DIFF
--- a/src/core/compiler/wasm/validate.go
+++ b/src/core/compiler/wasm/validate.go
@@ -186,6 +186,12 @@ type moduleValidator struct {
 	compCache       map[uint32]compCacheEntry
 	compCacheFrozen bool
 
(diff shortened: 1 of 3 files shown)


jtenner:

PR #530, opened 2026-08-28
runtime: validate interrupt linear-memory registers

jtenner:
Small correctness fix: a Linux interrupt signal can land in generated code before the entry prologue has installed the active linear-memory base in RBX/X26. The handler previously dereferenced that potentially stale register.

Bound `JobMemory` bases are now published in a fixed lock-free table. The assembly handlers require an exact registered base before reading the trap-cell pointer, and close disables an entry before unmapping. Cached mappings stay registered.

Proof:
- `go test ./src/core/runtime -run '^(TestBindTrapCellRegistersInterruptLinearMemory|TestJobMemoryHasTrapCellDetectsCrossInstanceOverwrite)$' -count=3`\n- `go test ./src/wago -run '^(TestCallContextInterruptsNativeLoop|TestRuntimeCloseInterruptsNativeLoop)$' -count=5`\n- `GOOS=linux GOARCH=arm64 go test -c ./src/core/runtime`\n\nFootprint: the allocation-free table is 32 KiB and supports 4096 concurrently bound memories.


jtenner:

jtenner:
@codex review

jtenner:
@codex review

PR #531, opened 2026-08-28
runtime: preserve chained signal semantics

jtenner:
Small correctness fix: installing the Linux guard-page handler replaced the predecessor’s signal mask and restart/defer behavior. On amd64, a non-Wasm fault inside a live reservation could also reach an obsolete framed fallback that dereferenced arbitrary RBP-relative memory.

The installer now creates distinct SIGSEGV/SIGBUS actions preserving each predecessor mask plus `SA_RESTART`/`SA_NODEFER`, and rejects one-shot predecessors that cannot be safely chained. The handler now recognizes only the current x64/WARP ABI and requires an exact pinned-RBX owner match before dereferencing basedata.

Proof:
- `go test -tags wago_guardpage ./src/core/runtime`
- `go test -tags wago_guardpage ./src/wago -run '^(TestMemoryBytesAfterGrowGuardPage|TestGuard|TestRuntime)'`\n- `GOOS=linux GOARCH=arm64 go test -c -tags wago_guardpage ./src/core/runtime`

### DIFF
--- a/src/core/runtime/sigtrap_amd64.s
+++ b/src/core/runtime/sigtrap_amd64.s
@@ -6,7 +6,7 @@
 // DX=*ucontext). Pure asm, no Go calls, no g use — runs on the signal alt-stack.
 // It derives everything per-fault (no per-call shared state): scan the live
 // reservation table for one containing the fault address, confirm the faulting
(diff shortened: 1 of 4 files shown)


jtenner:

jtenner:
@codex review

PR #539, opened 2026-08-28
wago: admit wide scalar GC host imports

### LINKED ISSUE #536: [P1][GC] Admit wide scalar host imports through the synchronous host ABI
## Summary

Exact native GC-root admission applies the small reference register-ABI check to scalar-only host imports when the module has no collector-reference import boundary.

A scalar import with more than two results is therefore rejected even though it transfers no GC references and fits the synchronous host-call ABI.

This disables collection for the Dewdrop self-hosted compiler.

## Environment

- Platform: Linux/AMD64
- Core features: v3

## Reproduction shape

The failing Facet import is scalar-only:

```wat
(type (func
  (param i32)
  (result i32 i32 i64 i64 i32 i64 i32 i64 i32 i32)))
```

The module also contains normal WasmGC allocations, so exact native roots are required.

Admission reports:

```text
GCNativeRootAdmission {
  Required: true,
  Exact: false,
  Reason: "a function import or caller exceeds the exact native call ABI",
}
```

Temporary diagnostic instrumentation identified function import 7.

## Cause

In `newGCFrameRootPlan`:

```go
collectorBoundary := moduleHasCollectorReferenceCallBoundary(m)
...
if collectorBoundary {
    gcFrameHostCallABI(ft)
} else {
    gcFrameReferenceCallABI(m, ft)
}
```

`moduleHasCollectorReferenceCallBoundary` is false because every import is scalar-only. That selects `gcFrameReferenceCallABI`, which rejects `len(ft.Results) > 2`.

The actual import is a synchronous host call. Its ten scalar result slots fit `gcFrameHostCallABI` and do not need collector-reference transfer handling.

## Expected behavior

Admission should classify each call boundary by its actual lowering and ownership needs. A scalar-only synchronous host import should use the checked synchronous host ABI even when no import transfers collector references.

The collector-reference-boundary flag should not force unrelated scalar host imports through the register-call ABI.

## Acceptance criteria

- Add a WasmGC module with a scalar-only host import returning at least three scalar results.
- `GCNativeRootAdmission()` is exact when the signature fits the synchronous host-call capacity.
- Collector-reference imports still require exact reference ownership and root handling.
- Dynamic typed-reference calls keep the smaller register-ABI proof.
- Add equivalent AMD64 and ARM64 coverage.

jtenner:
## Summary

- classify reference-free function imports through the checked synchronous host ABI on AMD64 and ARM64
- keep collector-reference ownership/root handling tied to the existing collector-boundary proof
- keep dynamic `call_ref` on the smaller register-ABI proof
- avoid forcing reference-free cross-instance calls to join the provider collector domain
- add the reported ten-result scalar import regression and an explicit three-result dynamic-call counter-test

## Why

Exact native GC-root admission previously selected the internal reference-call ABI for every function import whenever the module had no collector-reference import boundary. That rejected a reference-free synchronous import after two results even though the actual parked host ABI accepts up to 64 slots in each direction.

The newly admitted Wasmtime `gc/issue-13152` consumer also exposed a downstream ownership bug: preferred-domain selection treated every `InstanceExport` as a collector edge. Its reference-free imported function therefore tried to join a collection-disabled provider to a collection-enabled consumer. Reference-free calls transfer no collector ownership, so the instances may safely retain independent domains and GC policies.

This change classifies only reference-free imports through the synchronous capacity and ignores those imports during preferred collector selection. Reference-bearing functions plus imported GC globals/tables retain existing domain ownership, while dynamic typed-reference calls remain bounded by the register ABI.

## Verification

- `go test ./src/core/... -count=1`
- `go test ./src/wago -run "^TestWasmtimeCore3Corpus$" -count=1` — 215 modules and 690 assertions, zero skips
- direct `gc/issue-13152` fixture replay — 2 modules and 1 assertion, zero skips
- `go test ./src/wago -run "^(TestGC|TestPluginGC)" -count=1`
- focused explicit and `wago_guardpage` ABI tests
- focused race-detector tests
- `GOOS=linux GOARCH=arm64 go test -c -o /tmp/wago-issue536-arm64.test ./src/wago`

A full `go test ./src/wago -count=1` was also attempted locally. Its pinned official-spec cases stop before execution because the host has WABT 1.0.36 while the repository requires 1.0.41; the checked-in Wasmtime Core 3 corpus and all non-WABT suites above pass.

## Performance and footprint

No generated hot path, runtime layout, ABI capacity, or persistent metadata changes. Both changes are bounded compile/instantiation classification with no new allocations, so no benchmark or memory delta is expected.

## Docs

Unchanged: these internal correctness fixes do not alter developer workflow, public configuration, or supported ABI limits.

Fixes #536

### DIFF
--- a/src/wago/gc_frame_roots.go
+++ b/src/wago/gc_frame_roots.go
@@ -101,6 +101,23 @@ func wasmFuncTypeTransfersCollectorRefs(m *wasm.Module, ft *wasm.CompType) bool 
 	return false
 }
 
+func wasmFuncTypeReferenceFree(ft *wasm.CompType) bool {
+	if ft == nil {
+		return false
+	}
+	for _, typ := range ft.Params {
+		if typ.Kind() == wasm.ValRef {
+			return false
+		}
+	}
+	for _, typ := range ft.Results {
+		if typ.Kind() == wasm.ValRef {
+			return false
+		}
+	}
+	return true
+}
+
 func moduleHasCollectorReferenceCallBoundary(m *wasm.Module) bool {
 	if m == nil {
 		return false
--- a/src/wago/gc_frame_roots_arm64.go
+++ b/src/wago/gc_frame_roots_arm64.go
@@ -45,7 +45,11 @@ func newGCFrameRootPlan(m *wasm.Module, exactRoots bool) *shared.GCModuleFrameRo
 			if !ok {
 				return reject("function import %d has no validated signature", funcImport-1)
 			}
-			if collectorBoundary {
+			// Reference-free imports use the parked synchronous wrapper and need
+			// not also fit the internal register ABI. Preserve the existing
+			// collector-boundary path so reference ownership and root handling
+			// remain coupled to that module-level proof.
+			if collectorBoundary || wasmFuncTypeReferenceFree(ft) {
 				if !arm64GCFrameHostCallABI(ft) {
 					return reject("function import %d exceeds the synchronous host-call ABI", funcImport-1)
 				}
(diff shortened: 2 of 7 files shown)


jtenner:
@codex review

jtenner:
@codex review

PR #541, opened 2026-08-28
gc: preserve mixed native allocation extents

### LINKED ISSUE #538: [P1][AMD64] Investigate gc-native-alloc cast failure under collection
## Summary

With exact native roots admitted and normal Throughput collection enabled, the Dewdrop self-hosted compiler traps with `cast failure` during request decoding when `gc-native-alloc` is enabled.

Disabling only `gc-native-alloc` avoids the failure for at least 20 minutes on the same artifact and GC configuration.

This report is not yet minimized, but the A/B result is strong enough to track as an AMD64 GC correctness issue.

## Environment

- Platform: Linux/AMD64
- Core features: v3
- Collector: Throughput
- Heap: 512 MiB
- Nursery: 64 MiB

## Reproducer artifacts

Compiler module:

```text
size:   2326752 bytes
sha256: d22ecd466084b553eeef05ca1ccdb3f498210e98b69e24b444657c1c0b5aab56
```

Compile request:

```text
size:   9734989 bytes
sha256: a2acdeacccd2be3b4cd8fd9c881455ae58d60c71259f62cb24682af211be0d43
```

I can provide both artifacts or add a reduced repository fixture.

## Required admission setup

The stock runtime first hits three separate capacity/admission blockers. They are tracked in #535, #536, and #537.

For this A/B test, both runs used the same local diagnostic changes:

1. synchronous host-call capacity raised from 64 to 512 slots;
2. scalar host imports admitted through the synchronous host ABI;
3. `GCFrameRootLimit` raised from 1,024 to 2,048.

The resulting plan was exact:

```text
ADMISSION {
  Required: true,
  Exact: true,
  Safepoints: 19078,
  Callsites: 26812,
  MaximumRoots: 62,
  MetadataBytes: 1179633,
}
```

These same admission changes were present in both runs below. The only compiler-option difference was `gc-native-alloc`.

## Result with `gc-native-alloc` enabled

The default configuration traps after about 14 seconds:

```text
wasm trap: cast failure
    at decode_self_host_compile_request (func[332])
```

## Result with `gc-native-alloc` disabled

Configuration delta:

```go
cfg := wago.NewRuntimeConfig().
    WithCoreFeatures(wago.CoreFeaturesV3).
    WithOptimization("gc-native-alloc", false)
```

The same workload ran for at least 1,200 seconds without the cast trap. It was externally stopped by the test timeout.

As another control, collection-disabled execution reached heap exhaustion after 2,606 seconds with a 3 GiB heap and did not produce the cast failure.

## Working hypothesis

A native allocation fast or fallback path fails to preserve, publish, or reload a live reference across collection. The later `ref.cast` observes a reclaimed or wrong-typed object.

The admission changes are a caveat, but they are identical in the enabled/disabled comparison.

## Acceptance criteria

- Reduce this to a collect-every-allocation AMD64 regression.
- Run it with `VerifyAfterCollect` under Throughput and Tiny where applicable.
- Prove constructor inputs, unpublished objects, retained handle batches, and helper results remain exact roots across fallback collection.
- Keep `gc-native-alloc` enabled without the cast failure.
- Add a control that fails before the fix and passes with the optimization disabled.

jtenner:
Fixes #538

## Summary

- add a minimized native-allocation enabled/disabled regression for the reported ref.cast failure
- preserve the highest live Eden extent when compacting nursery handles
- retain collector layout, native allocation fast paths, and zero-allocation cleanup

## Root cause

Native arrays reserve 4 KiB chunks while native structs allocate directly from the same reservation batch. Their handle publication order can therefore differ from physical Eden address order. Collection assumed the final retained handle was the highest live object and could rewind nurseryBump below a live struct; a later helper fallback then overwrote the object and ref.cast trapped.

The fix computes the maximum live Eden extent during the handle-compaction scan that already visits every nursery handle.

## Validation

- go test ./src/wago -run TestGCNativeMixedReservationPreservesStructAcrossFallbackCollection -count=20
- go test -race ./src/wago -run TestGCNativeMixedReservationPreservesStructAcrossFallbackCollection -count=1
- go test ./src/core/runtime/gc -run TestNative\|Test.*Nursery\|Test.*Collection -count=1
- full src/wago package execution was attempted; self-contained tests passed, but staged official tests require the absent optional tests/spec-v3 checkout

## Performance and footprint

BenchmarkMinorCollectionCleanupScaling/old=0/survivors=0-of-64, five 500 ms samples: median 544.0 ns/op before and 541.0 ns/op after; 0 B/op and 0 allocs/op in both. No runtime-state layout or allocation hot-path change.

## Docs

No workflow or public API change.

### DIFF
--- a/src/core/runtime/gc/collect.go
+++ b/src/core/runtime/gc/collect.go
@@ -819,23 +819,21 @@ func (c *Collector) compactNurseryHandles() {
 	out := c.nurseryHandles[:0]
 	var edenMax uint32
 	if c.survivorBump == 0 {
-		// Eden is bump allocated and nurseryHandles preserves allocation order.
-		// Compact membership once, then find the last live Eden allocation from
-		// the tail instead of recomputing a maximum across every survivor.
+		// Native array chunks and direct native structs share one reserved handle
+		// batch. Their handle publication order can differ from their physical
+		// Eden order, so derive the bump from every live extent instead of trusting
+		// the last retained handle.
 		for _, h := range c.nurseryHandles {
 			if h == 0 || int(h) >= len(c.handles) {
 				continue
 			}
-			sp := c.handles[h].space
+			e := c.handles[h]
+			sp := e.space
 			if sp == spaceNursery || (sp == spaceLarge && c.handles[h].young()) {
 				out = append(out, h)
 			}
-		}
-		for i := len(out) - 1; i >= 0; i-- {
-			e := c.handles[out[i]]
-			if e.space == spaceNursery && e.off < c.edenBytes() {
+			if sp == spaceNursery && e.off < c.edenBytes() && e.off+e.size > edenMax {
 				edenMax = e.off + e.size
-				break
 			}
 		}
 		clear(c.nurseryHandles[len(out):])
--- a/src/wago/gc_native_mixed_alloc_amd64_test.go
+++ b/src/wago/gc_native_mixed_alloc_amd64_test.go
@@ -1,0 +1,68 @@
+//go:build linux && amd64 && !tinygo && !wago_guardpage
+
+package wago
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/wago-org/wago/src/core/compiler/wasm"
+	"github.com/wago-org/wago/tests/wasmtest"
+)
+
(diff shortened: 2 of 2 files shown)


jtenner:
@codex review

PR #543, opened 2026-08-28
wago: admit wide scalar GC callers

### LINKED ISSUE #536: [P1][GC] Admit wide scalar host imports through the synchronous host ABI

jtenner:
## Summary

- remove the stale AMD64 check that requires every native-call caller to fit the small reference-call register ABI
- stop applying the ARM64 imported-call ABI check to direct local calls, which already select the register or wrapper path
- keep the per-target ABI proof for dynamic typed-reference calls
- add an AMD64 and ARM64 regression with one live GC reference across a scalar host call that returns ten values

The wide local caller uses the wrapper ABI. The scalar import uses the synchronous host ABI. Both paths already publish exact native roots.

Closes #536

## Validation

Passed:

- go test ./src/wago -run "^(TestGC|TestPluginGC)" -count=1
- go test -tags wago_guardpage ./src/wago -run "^(TestGC|TestPluginGC)" -count=1
- go test ./src/wago -run "^(TestPluginGCWideScalarCallerUsesSynchronousABI|TestGCWideDynamicCallRefRetainsRegisterABIProof)$" -count=20
- go test ./src/core/compiler/backend/railshot/amd64 ./src/core/compiler/backend/railshot/arm64 ./src/core/compiler/backend/railshot/shared -count=1
- GOOS=linux GOARCH=arm64 go test -c -o /tmp/wago-issue536-arm64.test ./src/wago

The first CI run exposed and confirmed the ARM64 direct-local variant. The follow-up limits its imported-call proof to actual imports. Native ARM64 CI executes the shared regression.

The complete local go test ./src/wago -count=1 run reached unrelated staged-suite checks and failed because the local wast2json is 1.0.36. The repository requires 1.0.41.

## Performance and footprint

AMD64 removes one extra linear function-body scan. ARM64 keeps its existing scan and narrows one branch. The change adds no runtime operations, allocations, persistent metadata, or memory.

### DIFF
--- a/src/wago/gc_frame_roots_arm64.go
+++ b/src/wago/gc_frame_roots_arm64.go
@@ -275,12 +275,14 @@ func arm64GCFrameBodySafe(m *wasm.Module, body []byte, classifier *wasm.ModuleIn
 			if !ok {
 				return false
 			}
-			if int(imm.Index) < m.ImportedFuncCount() && (collectorBoundary || wasmFuncTypeReferenceFree(ft)) {
-				if !arm64GCFrameHostCallABI(ft) {
+			if int(imm.Index) < m.ImportedFuncCount() {
+				if collectorBoundary || wasmFuncTypeReferenceFree(ft) {
+					if !arm64GCFrameHostCallABI(ft) {
+						return false
+					}
+				} else if !arm64GCFrameReferenceCallABI(m, ft) {
 					return false
 				}
-			} else if !collectorBoundary && !arm64GCFrameReferenceCallABI(m, ft) {
-				return false
 			}
 		case 0x11:
 			ft, ok := m.TypeFunc(imm.Index)
--- a/src/wago/gc_frame_roots_linux_amd64.go
+++ b/src/wago/gc_frame_roots_linux_amd64.go
@@ -162,9 +162,6 @@ functions:
 		if maximumLive > shared.GCFrameRootLimit {
 			return reject("function %d has %d simultaneously live collector locals, limit %d", function, maximumLive, shared.GCFrameRootLimit)
 		}
-		if !collectorBoundary && bodyUsesNativeCall(m.Code[function].BodyBytes, &classifier) && !gcFrameReferenceCallABI(m, ft) {
-			return reject("function %d exceeds the exact native caller ABI", function)
-		}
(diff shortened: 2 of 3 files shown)


jtenner:
@codex review

jtenner:
@codex review

PR #547, opened 2026-08-28
gc: remove repeated helper-boundary root scans

jtenner:
## Summary

Remove two O(heap/global-count) costs from ordinary WasmGC allocation helper boundaries.

The Dewdrop self-host compiler has about 2,991 reference globals and a large live nursery set. Before this change, a 90-second CPU profile spent 78% of samples synchronizing every global root at every allocating helper. After the first optimization, native-batch cancellation exposed a second full nursery compaction that consumed about half of samples.

## Changes

- Skip validation and write barriers when a persistent root slot already contains the requested reference.
- Defer module-global root synchronization until a collection actually enumerates the parked frame root set.
- Use one spare bit in the existing frame-layout byte, preserving `gcPublicState` size.
- Trim the exact 32-handle native reservation tail during cancellation.
- Retain full nursery compaction as a fail-closed fallback when the tail invariant does not match.
- Add dense-prefix correctness and benchmark coverage.

## Measurements

```text
BenchmarkGCPersistentRootBarrier
before: about 8.3 ns/op
 after: about 1.8 ns/op
```

```text
BenchmarkNativeAllocationBatchCancelDenseNursery
before: about 116 us/op
 after: about 250 ns/op
```

On the 2.3 MiB Dewdrop compiler artifact:

- before: `syncGenericGCGlobalRootsLocked` consumed 78% of a 90-second profile;
- after unchanged-slot fast path: native-batch compaction consumed about 50%;
- after both commits and lazy synchronization: neither path appears in the top profile entries, and external generated code becomes the largest sample.

The workload still has a separate late null-reference failure in Dewdrop module emission. This PR is limited to the measured runtime costs and does not claim that the full bootstrap now completes.

## Proof

```text
go test ./src/core/runtime/gc
ok github.com/wago-org/wago/src/core/runtime/gc

go test ./src/wago
ok github.com/wago-org/wago/src/wago

go test ./src/core/compiler/backend/railshot/amd64
ok github.com/wago-org/wago/src/core/compiler/backend/railshot/amd64
```

Docs are unchanged. This is an internal collector/runtime performance fix with no workflow change.

### DIFF
--- a/src/core/runtime/gc/native_alloc.go
+++ b/src/core/runtime/gc/native_alloc.go
@@ -219,10 +219,37 @@ func (c *Collector) discardNativeStructHandles() {
 		c.nurseryBump = rewind
 	}
 	if canceled {
-		// Remove canceled identities immediately. A failed collection can return
-		// before normal nursery compaction, and later handle reuse must not append
-		// a duplicate identity to the dense young set.
-		c.compactNurseryHandles()
+		// Reserved handles are appended as one tail batch and no Go allocation or
+		// collection can append another nursery identity before cancellation. Trim
+		// that bounded tail directly and retain only consumed live handles. Keep the
+		// full compaction as a fail-closed fallback if the invariant is not exact.
+		trimmed := false
+		if int(end) <= len(c.nurseryHandles) {
+			start := len(c.nurseryHandles) - int(end)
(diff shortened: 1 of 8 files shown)


jtenner:
@codex review

PR #554, opened 2026-08-30
perf: eliminate memarg allocations and import rescans

jtenner:
## Summary

- decode indexed multi-memory memargs directly into `InstructionImmediate`, removing one allocation per classified instruction
- replace all three quadratic imported-function metadata scans with linear import-section passes
- defer valid function-import diagnostic formatting to error paths
- add mixed-import function-namespace and signature-order regression tests
- repair stale resource-limit and survivor-policy benchmark fixtures without reducing 10,000-memory decode and iteration coverage
- add import-name shape watchpoints and document retained and rejected audit experiments

## Measured results

Linux/amd64, AMD Ryzen 7 8845HS, Go 1.24.4. Focused runs used `GOMAXPROCS=1` and `-cpu=1`.

### Indexed memarg classification

10,000 indexed memory64 loads:

| Metric | Before | After | Change |
|---|---:|---:|---:|
| Time | 287.6 us/op | 211.7 us/op | -26.4% |
| Go bytes | 42,183 B/op | 1,915 B/op | -95.5% |
| Allocations | 10,001/op | 1/op | -99.99% |

The immediate-free control remains allocation-free.

### Import-heavy compilation

Ten-sample `benchstat` comparison against `main`:

| Imports | Metric | Before | After | Change |
|---:|---|---:|---:|---:|
| 1,000 | Time | 2.452 ms | 0.394 ms | -83.94% |
| 1,000 | Go bytes | 393,560 B/op | 307,608 B/op | -21.84% |
| 1,000 | Allocations | 7,798/op | 4,054/op | -48.01% |
| 10,000 | Time | 217.278 ms | 5.361 ms | -97.53% |
| 10,000 | Go bytes | 6,364,072 B/op | 5,486,008 B/op | -13.80% |
| 10,000 | Allocations | 79,818/op | 40,071/op | -49.80% |

An adversarial follow-up profile found the first revision still spent 86.7% of samples in a remaining `FuncTypeIndex` rescan. The final revision removes that loop as well.

### Corpus compile guard

Ten interleaved one-iteration samples found no significant compile-time change. Allocation bytes and counts were identical in every sample.

| Corpus | Before | After | p value |
|---|---:|---:|---:|
| SQLite | 83.36 ms | 83.76 ms | 0.436 |
| Ruby | 847.5 ms | 844.8 ms | 0.796 |
| esbuild | 571.1 ms | 562.9 ms | 0.143 |

The compile-time geomean moved -0.43%.

### Release size

Linux/amd64 stripped release profiles:

| Profile | Before | After | Change |
|---|---:|---:|---:|
| manager | 8,601,784 B | 8,601,784 B | 0 B |
| runtime-standard | 8,466,616 B | 8,470,712 B | +4,096 B |
| runtime-minimal | 8,159,416 B | 8,163,512 B | +4,096 B |

## Measured and rejected

- adjacent import-module string reuse improved a synthetic repeated-name case, but regressed real SQLite and Ruby decode samples; reverted
- an inline mixed-memory width word removed one small allocation, but slowed construction from about 13.9 to 18.0 ns/op; reverted

## Validation

- `go test ./src/core/compiler/wasm ./src/core/compiler/frontend ./src/core/runtime/gc ./src/wago -count=1`
- complete `src/core/compiler/wasm`, `src/core/runtime/gc`, and `src/wago` benchmark sets with `-benchtime=1x`
- 10-sample import `benchstat` comparison
- 10 interleaved main/head corpus compile samples with `benchstat`
- `make lint`
- TinyGo `TestSyncHostBindingStaysCompact`
- Darwin/AMD64 `src/wago` test cross-build
- `go run ./tests/tools/docs-check`
- `git diff --check origin/main..HEAD`

The follow-up also repairs the resource-policy CI failures inherited from `main`: the quota fixture no longer converts a native `uintptr` back to a Go pointer, unsupported Darwin/AMD64 multi-memory execution skips through the existing Core 3 guard, TinyGo accepts its bounded 32-byte host-binding alignment, and release-size ceilings now track the measured products with about 389 KiB of standard-Go runtime headroom.

Full measurements and audit notes are in `docs/research/benchmark-audit-2026-08-30.md`.

### DIFF
--- a/OPTIMIZATIONS.md
+++ b/OPTIMIZATIONS.md
@@ -17,7 +17,21 @@ Legend: effort S/M/L · value ⬜ low · 🟦 medium · 🟩 high · ⭐ very hi
 
 ---
 
-## What's in place (updated 2026-08-29)
+## What's in place (updated 2026-08-30)
+
+**Benchmark-audit frontend wins (2026-08-30).** Indexed multi-memory memargs in
+allocation-free bytecode walks now decode directly into `InstructionImmediate`
+instead of constructing the AST pointer form. A 10,000-load mixed-width fixture
+improves **287.6→211.7 µs/op** (-26.4%), **42,183→1,915 B/op**, and
+**10,001→1 alloc/op**. Large function-import modules also avoid quadratic
+`ImportedFuncCount`/`FuncSignature` rescans: GC-boundary and synchronous-host-slot
+prepasses range the import section once, while frontend diagnostics format only
+on failure. The 10,000-import compile watchpoint improves **199.726→30.442 ms**
+(-84.8%), **6,364,072→5,486,008 B/op**, and **79,818→40,071 allocs/op**; sqlite3,
+ruby, and esbuild corpus compile medians remain flat with unchanged allocation
+counts. Stale resource-limit and survivor-policy benchmark fixtures were repaired.
+Adjacent import-name reuse and an inline mixed-memory width word were measured and
+rejected. See `docs/research/benchmark-audit-2026-08-30.md`.
 
 **Commutative self-updates and low-32 masks (2026-08-29).** AMD64 now
 accumulates every safe non-fixed `x = f(y) op x` form directly in `x` instead of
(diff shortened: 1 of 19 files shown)


jtenner:
@JairusSW @codex please review the benchmark methodology, retained performance changes, and rejected experiments.

jtenner:
@codex review

jtenner:
@codex review

jtenner:
@JairusSW The adversarial findings and inherited main CI failures are addressed. Codex is clean on the follow-up. Please approve when ready.

PR #559, opened 2026-09-03
Audit end-to-end correctness boundaries

jtenner:
## Summary

- tighten decoder and validator rules for references, packed GC fields, DataCount, branch casts, and atomics
- preserve exact feature, element, GC constant-expression, and artifact metadata
- fail closed for nil imports/invocation, memory close races, invalid mapping sizes, and disabled prepared calls
- implement the specified unshared-memory atomic wait/notify boundary and document it
- reject unsafe CLI signatures, version paths, duration overflow, and large semver ordering errors

The full audit record is in `docs/research/correctness-audit-2026-09.md`.

## Verification

- changed normal packages pass on Go 1.22.12
- changed core packages and focused public-runtime boundaries pass under `-race`
- `go test -tags wago_guardpage ./src/core/runtime ./src/wago`
- Core 2: 1,600 modules and 2,880 validation assertions passed
- Core 3 explicit and signal modes: 2,226 modules and 58,038 assertions passed with zero gaps
- exhaustive `wasm-tools` opcode-shape comparison: zero mismatches
- 1.6 million differential executions: zero semantic mismatches
- `go vet`, generation sync, formatting, documentation links, and six target cross-builds pass

The broad local test also found environment-only failures on the unchanged base:
Wine cannot create its lock in the restricted runtime directory, and the installed
TinyGo requires Go 1.25 or newer while the audit used Go 1.22.12.

## Performance

Go 1.22.12, `GOMAXPROCS=1`, repeated samples against current `origin/main`:

| Benchmark | Time delta | Allocations |
|---|---:|---:|
| Decode and validate | +0.94% | unchanged |
| Invoke one scalar function | -0.05% | unchanged, 0 B/op |
| Instantiate a small module | +0.73% | unchanged |
| Instantiate shared memory | +1.20% | unchanged |
| Analyze module requirements | +1.28% | unchanged |
| Scan a scalar function body | +0.18% | unchanged, 0 B/op |

No measured path added an allocation or showed a meaningful regression.


jtenner:
@codex review

PR #565, opened 2026-09-06
Reduce safe call and dispatch latency

JairusSW:
## Summary

- cache direct integer-entry metadata for ordinary `Instance.Invoke`
- specialize fixed-arity public and prepared calls without allocating
- propagate a compile-time proof for isolated immutable tables so safe `call_indirect` dispatch avoids global serialization
- skip empty reservation-map work and collapse duplicate lifecycle defer machinery
- retain the `entersyscall` / `exitsyscall` GC and scheduler boundary added by #562

## Performance

Apple M4 Max, Darwin/arm64, Go 1.26.5, `GOMAXPROCS=1`, `-cpu=1`, 12 alternating 500 ms samples, median, zero allocations:

| Benchmark | main | this PR | delta |
|---|---:|---:|---:|
| `BenchmarkExecCallOverhead_wago` | 30.80 ns | 30.96 ns | flat |
| `BenchmarkExec/tiny.add` | 34.01 ns | 30.04 ns | -11.7% |
| `BenchmarkExec/dispatch.apply` | 37.07 ns | 31.20 ns | -15.8% |
| `BenchmarkInvokeAddOne` | 115.40 ns | 42.71 ns | -63.0% |
| `BenchmarkPreparedInvokeAddOne` | 28.95 ns | 28.71 ns | flat |
| wazero raw-call reference | 23.47 ns | 23.56 ns | flat |

The low-level native boundary is intentionally unchanged. The gains remove Go-side dispatch, lifecycle, cache, and serialization overhead around it.

## Safety boundaries

- Native calls still pass through the scheduler handoff from #562.
- Direct public entry requires an isolated-entry proof, a private reference store, and no shared-native-control or dynamic/imported/store-owned GC domain.
- Runtime-owned reference stores retain the established admission and topology-lease path.
- Immutable-table isolation is compiler-proven; serialized/decoded modules conservatively discard the proof.
- Wrong-arity direct-call coverage verifies that invocation admission is released and the instance remains reusable.
- Trap handling, object `KeepAlive`s, close checks, argument narrowing, and result normalization remain on the direct path.

## Validation

- `go test ./... -count=1`
- `make test-guard`
- `make lint` (`staticcheck` unavailable locally; remaining lint gates passed)
- `(cd bench && go test . -count=1)`
- focused direct-entry and lifecycle suite with `-race`
- Linux/arm64, Linux/amd64, and Windows/amd64 `src/wago` test binaries cross-compiled

Docs unchanged: this is an internal hot-path optimization with no developer workflow change.

### DIFF
--- a/src/core/compiler/backend/railshot/amd64/compile.go
+++ b/src/core/compiler/backend/railshot/amd64/compile.go
@@ -1706,7 +1706,7 @@ func compileModuleWith(m *wasm.Module, opts CompileOptions) (*amd64.CompiledModu
 			fmt.Fprint(os.Stderr, ms.String())
 		}
 		keepCodeBuffer = true
-		return &amd64.CompiledModule{Code: code, CodeImage: codeBuffer, Entry: entry, InternalEntry: internalEntry, DirectPrepared: directPrepared, RequiresBMI2: requiresBMI2, RequiresAVX2: requiresAVX2, RequiresAVX512: requiresAVX512}, nil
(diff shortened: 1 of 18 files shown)


jtenner:
@codex please review.

PR #568, opened 2026-09-07
railshot: optimize regional residency and model version transitions

JairusSW:
## Summary

- add pointer-free residency telemetry and a bounded local-event tape
- run allocation-free aggregate and detailed transition planners in shadow mode
- report predicted local-version admissions, evictions, reloads, and dirty writebacks without letting the model affect generated code
- improve ARM64 signals-based BLAKE residency with a mode-aware lease budget while preserving explicit-bounds scratch capacity
- improve AMD64 explicit-bounds residency by reaching score-gated eviction at effective capacity and admitting R8 as a last-choice lease in call-free, control-free, bulk-memory-free regions
- record exact baselines, resource costs, safety gates, CI-discovered corrections, and all measured policy rejections

This is one combined review for the completed regional-residency and transition-model work. It is the first active slice of the broader Railshot R3 execution-performance plan; it deliberately does not claim that the remaining R3 lanes are implemented.

## Execution results

ARM64, Apple M4 Max, signals-based bounds:

- blake-as: 372,114 -> 366,308 ns/op (-1.56%)
- BLAKE3: 231,678 -> 228,548 ns/op (-1.35%)
- explicit bounds remains at the proven 18-lease ceiling

AMD64, Ryzen 7 7800X3D, Linux, pinned to CPU 2, explicit bounds:

- initial effective-capacity fix: blake-as -1.04%, BLAKE3 -0.30%
- last-choice R8 lease over that fix: blake-as -0.61%, BLAKE3 -6.13%
- two opposite-order 46-workload screens bracketed geomean at +0.11% to -1.05%; BLAKE3 remained 5.9-6.4% faster
- compile latency: +0.28% blake-as, +0.47% BLAKE3
- compiler heap and allocation counts: unchanged for the R8 slice
- module native bytes: +0.43% blake-as, +2.03% BLAKE3

## Transition model

The detailed model runs only when codegen statistics are requested. On ARM64 explicit bounds it predicts:

| Workload | Admissions | Evictions | Reloads | Dirty writebacks |
| --- | ---: | ---: | ---: | ---: |
| blake-as | 25 | 6 | 4 | 5 |
| BLAKE3 | 26 | 7 | 3 | 5 |

Ordinary compilation retains the cheaper aggregate shadow. Emitted native code and allocation counts are unchanged by the detailed model; measured retained heap grew by 16-28 bytes on the two target modules and ordinary compile-latency movement was noise-sized.

## Safety

- no unsafe runtime boundary or host-call fast path is introduced
- the event tape is pointer-free, worker-owned, capped at 32,768 events, and fail-soft
- both planners are shadow-only and cannot affect generated code
- RAX/RCX/RDX remain unavailable to AMD64 regional leases
- AMD64 R8 is last choice and is available only with explicit bounds inside the pre-existing call-free, control-free, inlining-free, EH-free, bulk-memory-free eligibility boundary; signals mode retains R8 as fixed scratch
- ARM64 keeps three transient registers: explicit bounds uses 18 leases because X27 holds the memory size; signals-based bounds can use 19 because X27 is free

## CI-discovered corrections

- ARM64: an early CI run caught `blake-as-simd.hashN` corruption in explicit mode at 19 leases. The final policy makes the limit bounds-mode-aware. The full explicit/signals differential corpus and `make test-guard` pass; the rejected explicit-mode measurements remain documented.
- AMD64: Windows guard CI caught `blake-as.wasm.hashN` corruption when signals mode leased R8. Native AMD64 Linux reproduced the wrong result. The final policy caps signals mode at nine leases while retaining the tenth R8 lease for explicit bounds.

## Rejected experiments

All experimental code was removed rather than left dormant behind flags:

- per-version final-read transfer: BLAKE3 about 14% slower
- Belady-style next-use eviction: blake-as about 18% slower, BLAKE3 about 46% slower
- clean-first eviction: no writeback reduction and more activation churn
- additive dirty-eviction cost, read cooldown, and dirty-definition admission credit: mixed/noise-sized results or excessive writeback churn
- direct replay of aggregate transition benefit: blake-as kernel grew 5,308 -> 7,600 bytes with pressure misses 132 -> 679; BLAKE3 grew 5,560 -> 7,804 bytes with misses 166 -> 695
- ARM64 effective-capacity fallthrough: no change to either BLAKE kernel
- ARM64 explicit-bounds 19/20-lease configurations: insufficient transient register floor and semantic corruption
- AMD64 signals-mode R8 leasing: fixed-scratch conflict and semantic corruption

## Proof

- `go test ./...`
- `go test ./src/core/compiler/backend/railshot/...`
- `make test-guard`
- native Linux/AMD64 backend and semantic-corpus tests
- ARM64 and native AMD64 explicit/signals differential corpus tests
- TinyGo and cross-platform CI on the constituent exact SHAs
- paired focused and full-corpus benchmarks documented in `research/railshot_exec_v3/current_main_residency_baseline.md`


JairusSW:
Added ARM64 straight-line memory-kernel optimization in e85a6a2c.

What changed:
- retain eight bounded source-specific bounds certificates and range later scalar loads only across non-trapping integer work
- cache `memBytes` in backend scratch X17 for register-ABI, call-free, control-free, table-free, bulk-free regional leaves; reload it after entry/interrupt setup so callers do not preserve it
- free X27 for a nineteenth regional local without reducing the three-register transient floor
- expose both paths through optimization definitions and kill switches

Matched Apple M4 Max, Darwin/arm64, Go 1.26.5, GOMAXPROCS=1, 12 order-balanced paired samples (`-benchtime=750ms`):

| Workload | Wago median | wazero median | Wago / wazero |
| --- | ---: | ---: | ---: |
| blake-as.hashN | 382,348 ns/op | 383,835 ns/op | 0.9966 |
| blake3.blake3_hash | 233,751 ns/op | 247,371 ns/op | 0.9474 |

Against the previous branch head, a five-pair target sweep improved scalar blake-as 391,428 -> 382,375 ns/op (-2.31%) and blake3 hash 238,803 -> 233,262 ns/op (-2.32%). SIMD BLAKE, raytrace, and float were effectively unchanged.

Local proof:
- `go test ./...`
- `cd bench && go test ./...`
- `go test -race ./src/core/compiler/backend/railshot/arm64`
- `make lint` (staticcheck unavailable locally; all other lint/generation/vet/JS gates passed)
- Linux AMD64 and ARM64 backend test binaries cross-compiled

JairusSW:
Added a broader ARM64 memory-loop optimization in 4e8e0b05.

- Caches the stable cancellation-cell pointer only in interruptible, call-free memory loops with spare residency and the full transient-register floor intact.
- Keeps every cancellation/close poll and every trap-cell read; no unsafe boundary is added.
- Preserves loop layout and native size by replacing the repeated dependent metadata load with an equal-size NOP.
- Adds the WAGO_ARM64_NO_LOOP_TRAP_CELL kill switch, schema registration, pressure/layout tests, and an end-to-end cached-memory-loop cancellation test.

Apple M4 Max, 7 alternating on/off process pairs, 500 ms per row:
- memory.sum: 318.6 -> 310.4 ns/op (-2.57%)
- sieve.count: 102,206 -> 101,521 ns/op (-0.67%)
- BLAKE-AS: -0.34%
- CoreMark: +0.03%

Validation:
- go test ./...
- benchmark module: go test ./...
- cached memory-loop cancellation test, 3 repetitions
- backend pressure/layout tests, 10 repetitions

PR #576, opened 2026-09-09
railshot: improve cross-corpus execution

JairusSW:
## Summary

- accelerate prepared integer calls and bounded immutable-table dispatch
- add general single-pass AMD64 loop, register-residency, canonical-i32, and memory hot-path improvements
- extend ARM64 prepared-call, SIMD-result, bounded memory-copy, loop-hot scalar-merge, declared-i32-local, and reused-address-parameter lowering
- keep every optimization architecture- or semantics-gated; there are no module names, hashes, corpus identities, benchmark-specific predicates, or unsafe opt-in mode

## Review fixes

The branch now addresses all four findings from the review of `5e789efe`:

1. The AMD64 prebound trampoline republishes its exact trap reentry address on every invocation. An end-to-end subprocess test prepares a trapping division, performs an ordinary floating-point invocation on the same instance, and then requires a normal division-by-zero Wasm trap.
2. The module-global regional lease uses a dedicated frame slot outside the canonical operand/result area, and frame elision is disabled while that lease exists. Execution tests require the optimization to fire for both one- and two-result memory-touching functions and prove the pinned global remains unchanged.
3. Interruptible functions exclude `RSI`, the interrupt-poll scratch register, from persistent wide-constant residency. An execution test compares the optimization enabled and disabled under register pressure with `Interruptible: true`.
4. Bounded prepared-entry resolution summarizes the immutable target bitset once per fixed-point iteration. The 1k/2k/4k fanout benchmark is now approximately 19/37/74 microseconds, versus the reviewed quadratic 3.6/13.9/55.0 milliseconds.

The branch now merges current `main` at `3ec74f1d`. The semantic conflict resolution preserves main's dirty value-pin persistence on trap exits and restores the borrowed module-global register before that persistence runs.

## Performance

Fresh AMD64 results at rebased head `afb7a5f0`: six alternating samples per engine and row, 300 ms, signal-backed bounds, `GOMAXPROCS=1`, Ryzen 7 7800X3D CPU 7.

| Metric | Result |
|---|---:|
| execution geomean vs wazero, 46 rows | **1.547380x** |
| execution wins vs wazero | **38/46** |
| compile geomean vs wazero, 42 rows | **3.722209x** |
| fewer compiler bytes/op vs wazero | **11.152661x** |

An exact alternating comparison against the reviewed unsafe head measured the corrected build at 0.996838x of that head's execution geomean: a 0.32% aggregate cost. The required per-call trap publication is visible primarily in the smallest boundary microbenchmarks; compute-heavy application rows are effectively unchanged.

The last qualified Apple M4 Max measurement before the review-only AMD64 fixes was 1.453301x vs wazero across the same execution corpus. ARM64 performance was not rerun after rebasing onto `709494c3`, so it is not presented as a current-head measurement.

## Validation

- `go test ./...` on Apple ARM64 after rebasing
- complete AMD64 backend and runtime suites on Linux AMD64
- focused regression tests under the race detector on Linux AMD64
- Linux and Windows AMD64 test-binary cross-compiles
- six-pair full AMD64 execution corpus with signal-backed bounds
- three-sample full AMD64 compile/resource corpus
- `git diff --check`

Docs: no developer workflow changed.


JairusSW:
Addressed all four findings and rebased onto current `main` (`709494c3`). New head: `afb7a5f0`.

1. **Trap context:** the prebound AMD64 trampoline now republishes the exact `stackTop-40` reentry address on every call. The new subprocess regression reproduces the sequential prepare -> ordinary floating-point call -> trapping prepared division sequence and requires a normal division-by-zero Wasm error. It SIGSEGVed before the fix and passes now.
2. **Module-global lease:** the saved register now occupies a dedicated frame slot beyond locals, never canonical result/spill slot zero. Frame elision is also forbidden while the lease exists. New one- and two-result execution cases assert that the optimization fired, validate the results, and prove the global remains unchanged. The one-result case changed the global to 210 before the fix and passes now.
3. **Interrupt poll clobber:** `RSI` is excluded from wide-constant residency whenever compilation is interruptible. The new pressured loop test compiles with `Interruptible: true`, asserts the optimization fired, and compares enabled/disabled execution over multiple iteration counts. Enabled returned zero before the fix and now matches the disabled oracle.
4. **Resolver scaling:** immutable target safety/work/depth are summarized once per fixed-point iteration by scanning set bits, then reused by all indirect callers. On Ryzen 7 7800X3D, the final 1k/2k/4k benchmark medians are approximately 19/37/74 us, replacing the reviewed 3.6/13.9/55.0 ms quadratic shape.

The rebase overlapped main's new trap persistence. The resolution deliberately restores the regional lease first and then runs main's `storeGlobalPins(..., true)`, preserving both invariants.

Validation on rebased head:

- full `go test ./...` on Apple ARM64
- full AMD64 backend/runtime and focused subprocess tests on Linux AMD64
- focused race tests
- Linux/Windows AMD64 cross-compiles
- AMD64 execution: 1.547380x geomean vs wazero over 46 rows, 38/46 wins
- AMD64 compilation: 3.722209x faster and 11.152661x fewer bytes/op vs wazero over 42 rows

Please re-review the new head.

JairusSW:
Merged current main at 1ea71409 and fixed the resulting test-helper relocation by updating all stale tests/wasmtest imports to tests/support/wasmtest. Post-merge validation is green: canonical make test, root go test ./..., bench module go test ./..., benchmark smoke coverage, full AMD64 backend/runtime, non-staged AMD64 Wago, and focused race tests. The local Ryzen scratch host lacks OCaml/dune for interpreter-backed staged Core 3 tests; the fresh CI matrix provisions and covers that lane.

PR #581, opened 2026-09-09
Fix JIT, invocation, validation, and v1 settings correctness

jtenner:
This fixes nine correctness issues in nine focused commits: a GP/XMM allocation error, lost global writes on traps, missed cancellation after queued calls, two validator typing errors, three settings compatibility/order errors, and HostExit pointer recovery.

## Changes

- Keep XMM values out of amd64 GP spill selection. Cover all XMM register numbers and scalar float, SIMD, and custom value classes with allocator invariant tests.
- Persist dirty value-pinned globals on generated amd64 and ARM64 trap exits. Entry-trap paths reload pins before writeback because the prologue has not initialized them. ARM64 shared trap metadata uses registers outside the pin pool. Cover i32/i64 global persistence, both calling conventions, and shared/ordinary trap emitters.
- Check context cancellation after acquiring the instance gate on raw and host-token invocation paths, before export lookup or execution.
- Constrain polymorphic `ref.as_non_null` and `br_on_null` results to references using a validator-only heap bottom type. Preserve declared `try_table` results when its body is unreachable in both AST and streamed validation. Correct two exception fixtures that relied on the old invalid typing rule.
- Apply feature disables after enables so conflicting extended-constant settings resolve consistently. Accept retired v1 optimization names in global settings and restore `inline-loop-callees` and `deep-fp-pins` to the compatibility list and generated schema.
- Convert non-nil `*HostExit` panics to `ExitError` in synchronous native host calls and host-log replay, matching direct host re-exports.

Each fix includes focused tests and an architecture documentation update. Tests and fixes are combined per issue to keep each commit bounded and reviewable. Codex was used to inspect the code, implement the fixes, write tests/docs, and run validation.

## Validation

CI on `f4490418376ccda6852df4d31fd7662f8ff57a9b`: all 27 checks passed, including the native platform suites, conformance, race detector, TinyGo, and build-size checks. The requested `@codex` review completed on this commit with no findings. No new Markdown files are included; only the existing `ARCHITECTURE.md` is updated.

Local checks passed:

- `go test ./src/core/compiler/wasm -count=1`
- `go test ./cli/internal/settings ./cli/internal/project -count=1`
- Focused API and amd64 backend tests for register allocation, traps, globals, host exits, cancellation, context propagation, and staged exception execution.
- `go test -race ./src/wago -run '^TestQueuedInvocationChecksCancellationAfterGate$' -count=20`
- `go vet ./src/wago ./src/core/compiler/wasm ./src/core/compiler/backend/railshot/amd64 ./cli/internal/settings ./cli/internal/project`
- Linux/ARM64 API and backend test binaries cross-compile.

The official exception-family accounting test was rerun with WABT 1.0.41 installed through mise. `throw` and `throw_ref` passed 21 assertions. The overall test still fails on setup: WABT cannot parse the `tag`, `try_table`, and `ref_null` fixtures, and the required pinned Release 3 interpreter is not configured through `WAGO_SPEC_INTERPRETER`. The accounting snapshot was not updated to hide this failure.

The full test/benchmark corpus was not run locally, as requested. Local ARM64 binaries were not executed because QEMU is absent; the native ARM64 CI suites subsequently passed. Native ARM64 performance, native-code sizes, and peak resident memory were not measured.

## Performance

Compared against `7b3efcedaa674c8dd05e91909daa0138476f107b` on Linux/amd64, Ryzen 7 8845HS, Go 1.27.1, 16 Go processors, default build flags. An initial benchmark ran before edits. After a workspace reset cleared its temporary captures, the unchanged starting revision was measured again in a separate worktree.

All values below are medians of six samples. Small-module public compile rows use a 300 ms confirmation run with candidate measured before baseline. Other rows use 100 ms samples with baseline measured before candidate. Root-package 100 ms runs used five samples plus one additional sample. Positive change means slower.

| Benchmark | Before, ns/op | After, ns/op | Change |
|---|---:|---:|---:|
| CompileSmallScalar | 16,498 | 16,360.5 | -0.83% |
| CompileSmallScalarCore3 | 17,963.5 | 17,499.5 | -2.58% |
| RailshotCompileSmallScalar | 6,564.5 | 6,593.5 | +0.44% |
| RailshotCompileMediumControl | 6,986 | 6,964.5 | -0.31% |
| RailshotCompileSIMDHeavy | 8,046.5 | 7,889.5 | -1.95% |
| Compile_wago | 10,037.5 | 10,081.5 | +0.44% |
| InvokeAddOne | 20.62 | 20.07 | -2.67% |
| InvokeNumericGlobalRoundTrip | 78.115 | 78.69 | +0.74% |
| InvokeHostFuncDirect | 488.2 | 486.7 | -0.31% |
| ExecFibLoop_wago | 27.685 | 27.8 | +0.42% |
| ExecFibRec_wago | 350,201 | 351,603 | +0.40% |
| ExecCallOverhead_wago | 21.87 | 21.66 | -0.96% |
| ExecHostRoundtrip_wago | 479.25 | 487.3 | +1.68% |
| ExecGlobalGet_wago | 75.46 | 75.655 | +0.26% |
| ExecGlobalSet_wago | 75.395 | 75.495 | +0.13% |
| ExecLocalGet_wago | 76.615 | 77.095 | +0.63% |
| ExecMemoryLoad_wago | 76.005 | 76.005 | 0.00% |

No large slowdown appeared. The largest increase was 8.05 ns (+1.68%) for a host-call round trip. Allocation counts are unchanged in all 17 cases. Host calls remain at 112 B/op and one allocation; other measured execution cases have zero allocations. SIMD backend compilation remains at 11,128 B/op and 24 allocations. Public scalar compilation remains at 15,000 B/op and 68 allocations, and Core 3 at 15,258 B/op and 77 allocations in the confirmation runs.

These small timing changes are subject to machine load and measurement noise; they are not claims of statistical significance. Raw captures are retained locally in the ignored `.perf/correctness-2026-09-09/` directory, per the benchmark documentation.


jtenner:
@codex review Please review the nine correctness fixes, with extra attention to register banks, trap pin persistence on both targets, cancellation after queueing, and validator stack types.

JairusSW:
within noise. perf is fine

PR #588, opened 2026-09-10
Reduce synchronous host-call overhead with instance-local state

jtenner:
## Summary

Reduce synchronous Wasm → host → Wasm overhead without changing scheduler admission, execution ownership, collector leases, parked roots, cancellation or callback authority.

The first pass makes activation/reservation bookkeeping instance-local, propagates trusted root identity, compacts immutable callback tokens, and skips redundant private native-context reconstruction. This follow-up adds:

- A compatible optional `CallerHostFunc` / `Caller` API and plugin `CallerFunc` builder. Direct concrete dispatch does not box the token. Existing `HostFunc` remains supported.
- A no-GC suspension branch proven from the **actual lease owner's** collector/domain metadata. Imported, dynamic and uncertain configurations retain general suspension.
- Once-published scope-pointer reuse, inline-first reservation lookup, and an activation-local root-context cache. Nested entries have distinct caches; callback generations remain fresh.
- The cross-platform scheduler fixture fix: basic admission assertions always run; the additional multi-memory fixture explicitly enables its feature.

Caller values retain private immutable generation snapshots and are never pooled. Both representations share helper authorization. Tests cover retained A during B, nested restoration, live externref denial, A → host → B → host → A, exact GC results, cancellation, exceptions and GC while parked.

**Direct control-frame scalar buffers and bounded scheduler segments remain disabled.** The complete native-stack continuation/progress proof is still missing. Arbitrary host code continues to run through the normal Go scheduler protocol.

## Measured checkpoints

Linux/amd64, Ryzen 7 8845HS, Go 1.27.1, GOMAXPROCS=16. Five 500 ms samples per main benchmark; small reservation changes and GC fallback comparisons also use ten 2-second samples. Matched host/guest controls use the same compiled host-capable export and public API at counts 0, 1, 8, 64 and 1024, with 0/1/4 memories and independent-instance parallel runs.

Bytes/allocations below are normalized per callback. Repeated cost subtracts the matched guest control.

| Stage | ns/call | B/call | allocs/call |
|---|---:|---:|---:|
| Fresh PR #588 baseline | 227.1 | 64 | 1 |
| Concrete caller | 159.6 | 0 | 0 |
| No-GC suspension | 139.3 | 0 | 0 |
| Scope/state cleanup | 124.9 | 0 | 0 |
| Reservation lookup | 132.1 | 0 | 0 |
| Activation-local context | 115.6 | 0 | 0 |
| Direct frame experiment | Not enabled | — | — |
| Final combined sweep | 115.8 | 0 | 0 |

These are sequential checkpoints, not additive attribution. The API-stage matched legacy/concrete comparison is 184.0 versus 159.6 ns/call. Reservation lookup improved its focused inline case from 5.26 to 3.61 ns, but did not show an isolated full-loop win. An initial split-helper GC layout was rejected; the accepted no-GC stage had an imported-domain tradeoff that later context caching removed.

Final public one-call medians: **353.3 ns concrete** (352.1–359.4), **395.8 ns legacy** (393.9–402.9), versus baseline 416.4 ns. Legacy still allocates 64 B/one object. Concrete runtime dispatch allocates zero after setup; host code may still allocate. Fixed zero-host invocation cost rose from 124.2 to 136.4 ns and is reported separately.

Parallel concrete throughput: **64.6 million callbacks/s** across 16 independent instances; legacy 29.6 million/s. Final concrete GC-capable loops take 215,934 ns local GC, 171,994 ns imported-domain and 186,272 ns dynamic-domain per 1,024-call invocation, all with zero Go allocations after setup.

The final ten-by-two-second paired fallback check confirms improvements over the concrete-API reference: local GC -12.2%, imported domains -7.6%, dynamic domains -9.4%. A separate parallel profile has no sampled contention on Wago activation locks.

The final profile is now led by unresolved external code (14.00% flat), scalar dispatch (8.88%), activation dispatch (8.43%), scheduler exit (6.40%), and native entry (5.19%). Generic no-GC suspension and repeated context/reservation lookup are absent from the private hot profile. Cumulative percentages are not added together.

Reports: [follow-up invariants, measurements, security review and tests](https://github.com/wago-org/wago/blob/perf/synchronous-host-roundtrips/docs/host-caller-performance.md), [all benchmark medians and ranges](https://github.com/wago-org/wago/blob/perf/synchronous-host-roundtrips/docs/host-caller-benchmarks.csv), [first-pass report](https://github.com/wago-org/wago/blob/perf/synchronous-host-roundtrips/docs/host-roundtrip-performance.md).

## Validation

- Full `src/wago` and race suites passed at every production checkpoint.
- Final standard-Go tests include 16 independent concrete-caller instances and stale-token denial against a live later callback's externref.
- Final `go test -race -count=1 ./src/wago ./src/core/runtime ./tests/integration/runtimeconcurrency` passed (`GORACE=atexit_sleep_ms=0` removes only process-exit delay).
- Guard-page tests, the separate benchmark/semantic-corpus module, all four bounded fuzz gates, and 122,993 codec fuzz executions passed.
- Escape analysis, generated facade and documentation checks passed.
- The [full final-runtime CI matrix](https://github.com/wago-org/wago/actions/runs/34478366701) and [final test/report matrix](https://github.com/wago-org/wago/actions/runs/34480083220) passed, including Darwin/amd64, both Windows targets, TinyGo, Linux/arm64, race, concurrency and conformance.

The full local repository suite still reports three external-tool failures: Wine cannot download its installer; two standalone TinyGo tests fail linking duplicate `tinygo_task_exit`. These are reported, not hidden or counted as passes. Native CI TinyGo lanes pass.

The final test/report matrix's first Linux attempt hit the unchanged CLI terminal job-control test; the job passed on retry. A local 20-repeat check had one log-wait timeout, while 20 reference-checkout repeats passed. The initial failure remains recorded in the report; all other matrix jobs passed on their first attempt.

AI assistance: Codex implemented the changes, tests, benchmarks and reports, reviewed the diff, and ran the recorded checks.


jtenner:
@codex review

Please review the latest changes, with focus on callback capability expiry, root versus callee invocation identity, activation-local context caching, and GC-domain lease safety. The latest CI and regression stress runs are green.

PR #606, opened 2026-09-11
Fix conditional snapshots with more than 16 pinned locals

jtenner:
Conditional paths must snapshot every pinned local. The combined integer and floating-point pin count can exceed a small inline buffer: dynamic function `ref.test` failed with 17 and 19 pins, and arm64 exception handling failed with 23 pins.

The amd64 struct and array stores retain their inline `[16]locState` buffer plus the existing pool for larger snapshots. Dynamic function `ref.test` and the dormant amd64 exception snapshot now use the same design. Pooled slices stay separate from inline arrays and return once after emission. Local-state reload rules and native fast paths are unchanged.

The audit also proved the same bug in arm64 `emitEHCatchRoute`. A separate commit replaces its 16-entry snapshot with the existing packed control-frame state format. This adds no allocator policy or compilation limit. Complete-slice snapshots and architectural register-bank candidate arrays need no change.

Regression coverage:

- Keep the original 19-pin struct/array test.
- Check actual pin counts of 16, 17, and 19 with no pin relinquishment. Change locals before the branch and compare separate results, including exact FP bits.
- Force an unremembered old parent to store a young heap-object child. Check all locals, the exact reference and payload, then clear the child root, collect explicitly, and check reachability.
- Use `wago_gcstats` to prove one helper fallback followed by 63 native stores with 19 live pins.
- Exercise dynamic `ref.test` through an exported mutable table: matching, unrelated, null, and a bare imported provider whose unknown compact type ID selects the Go helper.
- Check all four local states, active-snapshot isolation, and repeated pool returns in the amd64 emitter.
- Reproduce the arm64 failure before the fix, then compile and execute the 23-pin throw/catch case under qemu-aarch64.

The baseline documentation test now checks the existing `SPECTEST.md`, whose current totals match the committed baseline. Its validation logic is unchanged. No standalone Markdown files are added.

Validation passes:

- `go build ./...`
- `go test -count=1 ./src/core/compiler/backend/railshot/amd64`
- `go test -count=1 ./src/wago`
- `go test -count=1 ./tests/conformance/spectest`
- `go test -count=1 ./...` (73.30 s)
- Focused GC tests with `wago_gcstats`, including existing barriers and caller-saved pins.
- Focused arm64 control-state and exception tests under qemu-aarch64.
- Both saved Dewdrop map/SSA 17-pin reproducers return without a trap.
- Go 1.22.12 and 1.27.1 escape analysis keep all inline snapshots on the stack. A 16-pin dynamic function emits identical native bytes before and after this change.

Full validation used Go 1.22.12, CI's TinyGo 0.41.1, WABT 1.0.41, and the pinned Release 3 interpreter. `GOFLAGS=-buildvcs=false` avoids TinyGo VCS stamping failure in the linked worktree; no test was skipped or weakened for this repair. A larger synthetic-module stress framework remains a follow-up.


JairusSW:
hmm. should probably add some synthetic module with a heck ton of massive functions and locals. i've run into a similar problem where it panicked when there were too many pinned locals.
i'll do this in different pr though

PR #672, opened 2026-09-21
perf: reduce import setup allocations with measured lifecycle coverage

jtenner:
Command setup repeatedly creates import-key strings during snapshot validation. Read the callback already stored in each declaration instead, while retaining signature validation, locking, sealing, nil-handler errors, and instance isolation. This removes 46 allocations per command with a two-line production change.

The first commit contains that fix and its regression tests. The second contains the setup/cleanup diagnostics, corrected worker API/worker-count reporting and backend code-image cleanup, lifecycle/isolation tests, and the complete performance report with raw samples and reproduction commands. Existing end-to-end benchmark loops, compiler policy, memory policy, and dependency versions are unchanged. The large evidence commit is separate from the small production fix.

[Companion WASI PR #21](https://github.com/wago-org/wasi/pull/21) removes repeated private definition construction and an intermediate callback layer. It saves **another 70 allocations** per command. Wago's dependency pin is unchanged, so those additional gains require the provider change and a later dependency update.

| Controlled comparison | Result |
| --- | --- |
| Wago snapshot fix, 10 interleaved pairs | cjson -4.62%, tinyxml2 -3.00%; 46 fewer allocations each |
| Provider change, 20 interleaved pairs with snapshot fix in both builds | import construction -18.93%; cjson -10.28%; tinyxml2 -9.61%; 70 fewer allocations each |
| Direct host-call and host-call-heavy controls | No significant latency change; allocation counts unchanged |
| Compatible full baseline/candidate screens | 984 measurements passed in each; all 72 Wago execution cases at zero B/op and allocs/op |

The provider table has a one-time cost of 4,992 B and 24 allocations. Normal retained RSS did not change significantly; post-forced-GC tinyxml2 RSS increased 756 KiB (+3.56%, p=0.004). The full-screen process recorded 0 to 123 major faults and reached about 28.5 to 29.3 GiB peak RSS across all engines and workloads; these are not per-instance memory results. Twenty paired follow-ups did not confirm the six full-screen timing flags. No worker or memory production optimization was retained.

Validation includes unit, race, ownership, isolation, caller-mutation, descriptor-growth, fuzz, corpus/oracle, guarded-runtime, and host-call checks. Publication checks reran `just lint`, import unit/race tests, guarded Wago tests, affected benchmark race tests, provider vet/race tests, and documentation checks. Two TinyGo standalone tests still fail with duplicate linker symbols; the failure was reproduced on the new baseline. This is not an all-tests-pass claim, and another architecture was not tested.

See `docs/performance/setup-cleanup/README.md` for measurements, uncertainty, rejected hypotheses, source/dependency hashes, exact commands, and remaining limits. `PUBLICATION.md` records the publication bases and checks. Provider performance was measured at pinned revision `6a6684d2ecd2be2d17e5792733d1d0e03b2f2c0e`; its PR applies the identical core.go patch to current provider main and passed fresh vet/race checks, but that different dependency base did not receive a new full performance run.

Publication follow-up: the first provider CI run found a Go 1.22-only assertion error in the new closed-file test. The files were closed, but Stat's error did not match os.ErrClosed on that toolchain. A separate test commit checks the invalid descriptor directly. The original assertion failure was reproduced locally, and the complete corrected provider race suite passed on Go 1.22.12 and Go 1.27.1. Production code and measured benchmark loops are unchanged.


jtenner:
The focused retention work is ready for review. The accepted changes and tests are in separate commits. The full measurement report follows below; no new Markdown document was committed.

The evidence includes 858 original raw files in `raw-runs.tar.gz`, a per-file SHA-256 index, source overlays, build identities, summaries and rejected-candidate records. Extract the archive under the evidence directory to restore the recorded run directories. Historical raw files were preserved.

Publication checks: the new snapshot tests passed; provider `go vet ./...` and `go test -race ./...` passed with `GOWORK=off` and the existing pinned dependency. `just lint` passed before each commit. The repository's standard staticcheck step still reports its existing findings; its required runtime checks pass. An unused assignment in a new test was corrected. The optional Linux mapping observer is excluded from TinyGo because TinyGo lacks the required OS counters; no TinyGo runtime pass is claimed.

The publication checks used one CPU and heat controls. One lint attempt stopped at the 75 C threshold. A separate attempt failed because process pauses affected a 2-second script test. Both attempts are preserved. The final runner leaves only that short timer test unpaused. No performance samples or full-suite benchmarks were rerun for publication.

Wago's provider dependency pin is unchanged. The measured local workspace explicitly selected the provider PR. The input-copy experiment was rejected and is absent from production code. Normal sustained process-RSS reduction remains unproven.

Published Wago head: `b68c9254ebcd904bc9e99c5df0fbe259f6f63ac8`. Provider head: `641e1429b7617a5b7d50be3d35734cfc48dc56f2`.

[Measurement files, patches, and validation logs](https://github.com/wago-org/wago/tree/b68c9254ebcd904bc9e99c5df0fbe259f6f63ac8/docs/performance/setup-cleanup/retention-20260921).

Provider CI: all 14 checks passed on the new head. Wago CI is starting; this report does not claim it has passed.

Commits:

- [bfbe3d01d](https://github.com/wago-org/wago/commit/bfbe3d01d672751331074113f56cca5c7470b5cd): fix(imports): validate sealed snapshots only once
- [9dfed9ae4](https://github.com/jtenner/wasi/commit/9dfed9ae4e63945eddf2e05b6cb8608e93eecab1): perf(wasi): scan path depth without component allocations
- [ff69ad8e0](https://github.com/jtenner/wasi/commit/ff69ad8e07be38f7b102cfa628c189a3d8aab083): fix(wasi): release terminal descriptor map storage
- [7b179cc88](https://github.com/jtenner/wasi/commit/7b179cc88ea42eefc47073afa3d2d75e5d5a31dd): perf(wasi): detach state maps during provider shutdown
- [641e1429b](https://github.com/jtenner/wasi/commit/641e1429b7617a5b7d50be3d35734cfc48dc56f2): test(wasi): verify preopen ownership for argument-only instances
- [b644fccec](https://github.com/wago-org/wago/commit/b644fccec8e497432531691571a94eee1c58ffde): test(runtime): inspect input retention and native idle caches
- [b68c9254e](https://github.com/wago-org/wago/commit/b68c9254ebcd904bc9e99c5df0fbe259f6f63ac8): test(perf): archive focused retention measurement results

---

# Focused retention work — 21 September 2026

Four small fixes passed focused checks. One input-copy experiment was rejected. No normal-operation process-memory reduction is claimed. The measurements used isolated worktrees. Publication was authorized after measurement. No full 984-case suite was run, and no dependency pin was changed.

## Source and build identity

- Original Wago checkout: `/home/jtenner/Projects/wago`, branch `main`, HEAD `95be283fa511db7b01d86ef72b6c58fbe1ab607a`. It already had tracked and untracked work. Its tracked diff is preserved; see `baseline.json` and `preservation.json`.
- Accepted Wago reference: branch `perf/setup-cleanup-measured`, HEAD `cd26089d7d52ff6a8bd00712d3eb8c4734762b18` in `/tmp/wago-setup-cleanup-pr-uvyppyeh/wago`.
- Accepted provider reference: branch `perf/wasi-construction`, HEAD `aea44da450ec1e558a799a1b4f124bbbe355c250` in `/tmp/wago-setup-cleanup-pr-uvyppyeh/wasi-publish`. The requested `/home/jtenner/Projects/wasi` did not exist. This existing PR checkout was used instead.
- New detached worktrees: `/tmp/wago-retention-roobe8pc/wago` and `/tmp/wago-retention-roobe8pc/provider`. These contain the changes. The original checkouts and dependency pins were not replaced.
- The ordinary Wago bench pin is WASI `v0.3.1-0.20260916034125-6a6684d2ecd2`; it does **not** enable the provider PR. All task builds used the recorded `build.work`, with both local modules. `modules.jsonl` verifies the selected provider directory. No module-cache file was edited.
- Performance: Go 1.27.1, Linux amd64, GOAMD64=v1, `wago_guardpage`, `WAGO_BOUNDS=signals`, GOMAXPROCS=16, CPUs 0–15, GOGC=100, GOMEMLIMIT=off, empty GODEBUG. Ryzen 7 8845HS; kernel `6.12.107+deb13-amd64`. Build and machine records are included.
- Correctness checks used one CPU and throttled builds to control heat. Their times are not performance results. Go 1.22.12 checks are identified separately.

Each comparison includes the earlier accepted fixes on both sides. A source overlay removes only the change under test. The existing snapshot, construction, direct-dispatch, compact-definition, argument-callback and CodeImage-cleanup savings are not counted again. No combined improvement versus main is calculated from separate experiments.

## Measurements

Each final comparison used 20 matched fresh-process pairs, with alternating baseline/candidate order. The same diagnostic code was used on both sides. Timed leaves used 50 ms; fixed-work heap diagnostics used one operation. The corrected existing runner checks exact leaves, complete records, distinct output directories, build hashes, failures and heat. Benchstat output gives medians and confidence intervals; `p=0.000` means p<0.001.

An initial grouped path run reached the 75°C stop limit and was stopped. Its partial records remain in `path-pairs`; they were not used. Before final collection, the plan changed to one leaf per process, a start limit of 55°C, and a stop limit of 75°C. Final timing-run sampled peaks were at most 56.75°C. There was no search for favorable samples.

### 1. Path component scan — Implemented and verified

Source: provider `internal/core/fs.go:pathEscapes` and `resolve` (around lines 315 and 285). Patch: `01-path-scan-with-tests.patch`.

The old `strings.Split` made a component slice proportional to component count. The new scan uses substrings of the existing host string and an integer depth. It preserves empty components, dots, dot-dot, repeated/trailing separators and early escape. Its position relative to descriptor, directory, empty-path and absolute-path checks is unchanged. Path normalization, symlink handling and capability checks are unchanged.

| Operation | Before → after | B/op before → after | allocs/op before → after |
|---|---:|---:|---:|
| Depth check, short `a/b/c` | 32.740 → 9.921 ns | 48 → 0 | 1 → 0 |
| Depth check, 1,025 normal components | 7.807 → 4.752 µs | 18,432 → 0 | 1 → 0 |
| Depth check, 1,024 `./` prefixes | 7.788 → 4.794 µs | 18,432 → 0 | 1 → 0 |
| Full resolve, short | 468.8 → 440.3 ns (−6.07%) | 256 → 208 | 2 → 1 |
| Full resolve, many components | 11.481 → 7.873 µs (−31.42%) | 18,641 → 208 | 2 → 1 |
| Full resolve, normalizes to `b` | 9.982 → 6.731 µs (−32.57%) | 20,944 → 2,512 | 3 → 2 |

All timing differences above have p<0.001; median intervals are about 1–3%. Full resolve includes its existing descriptor Stat and path cleaning, but excludes guest-string construction. This is not a full-command timing claim.

Ownership/security: no new owner, cache, pointer alias, guest-memory reference or size limit. The depth scan allocates nothing. Tests cover the listed paths, invalid guest ranges and error order. Differential fuzzing compares with the old split algorithm; the first 3-second run completed 6,131 executions.

### 2. Repeated failed snapshots — Implemented and verified

Source: Wago `src/wago/imports.go:136`. Patch: `02-snapshot-validation-with-tests.patch`.

Repeated snapshots of a sealed invalid Imports object re-ran validation and appended equivalent errors with `errors.Join`. Validation now runs only before the existing sealed flag is set. Later calls return the current stored result or error. Rejected later mutations still call `record`, so they remain visible. There is no new cache or field.

One failed-batch operation creates an invalid import set and calls snapshot 256 times:

| Case | Before → after | B/op before → after | allocs/op before → after |
|---|---:|---:|---:|
| Nil event | 43.427 → 1.433 µs (−96.70%) | 43,683 → 808 | 1,544 → 14 |
| Signature mismatch | 122.532 → 1.866 µs (−98.48%) | 81,619.5 → 968 | 3,595 → 25 |
| Valid sealed set, 46 declarations | 42.450 → 5.321 ns (−87.47%) | 0 → 0 | 0 → 0 |

All timing differences have p<0.001, with about 1–2% median intervals. These failed-batch numbers are not per-command savings. The bounded tests prove stable error identity over 1,024 repeats and zero allocations on later failures.

Ownership/security: the same mutex protects sealing, validation, builders and errors. Tests cover complete first validation, nil and signature errors, valid reuse, concurrent reuse, and rejected mutation after success and failure. The other sealed constructor (`Instance.Imports`) contains validated bindings and no declarations. Distinct rejected API mutations can still add errors; that separate behavior was not changed.

### 3. Terminal descriptor maps — Implemented and verified

Source: provider `internal/core/fs.go:209`. Patch: `03-terminal-descriptor-map-with-tests.patch`.

`closeFS` closed owned descriptors and cleared the map, but retained the empty map allocation through the terminal state. It now sets `state.fds = nil` after the same close loop. `Plugin.fs` stays present while the provider is available. The claimed flag prevents reuse of that terminal first state; later instances get fresh state.

The synthetic create-and-close fixture changed from 237.7 to 226.7 ns for three descriptors (−4.63%) and 55.82 to 54.31 µs for 1,024 descriptors (−2.72%). Allocation traffic was unchanged. Both timing differences have p<0.001.

A separate fixed-work diagnostic kept eight terminal states alive, each after 4,096 map entries. Post-GC HeapAlloc fell from 1,546.7 to 391.9 KiB: about **1,154.8 KiB of retained Go heap removed** (p<0.001). This isolates map storage, with one shared empty entry per map. It does not model 32,768 live OS descriptors. Normal pre-GC HeapAlloc was 1.509 versus 1.507 MiB (p=0.644); no normal-memory improvement is established by that endpoint. Forced-GC diagnostic time is not production latency.

Ownership/security: cleanup only closes owned files and directory iterators. Borrowed standard streams are unchanged. Terminal state cannot accept new descriptors. Tests include real open files, provider reuse, normal exit, traps, start traps, partial setup, cancellation, repeated Close, independent concurrent instances, and bounded descriptor counts. No forced GC is used to establish correct ownership.

### 4. Provider shutdown map — Implemented and verified

Source: provider `internal/core/fs.go:237`. Patch: `04-provider-shutdown-with-tests.patch`.

Before editing, the ownership graph showed that the initial `Plugin.fs` can alias at most one entry in `guard.states`; all other states have distinct instance identities. `stateFor` is the only insertion path. Shutdown now detaches the state map and initial state under the existing guard lock, sets the closed marker, then closes states outside that lock. It skips the one possible initial-state alias. The temporary deduplication map is no longer needed. The detached map dies when shutdown returns.

| Synthetic state construction plus shutdown | Before → after | B/op | allocs/op |
|---|---:|---:|---:|
| One state | 446.5 → 367.2 ns (−17.78%) | 872 → 872 | 9 → 9 |
| Sixteen states | 4.810 → 4.268 µs (−11.26%) | 11.41 → 10.80 KiB | 104 → 101 |

Both timing differences have p<0.001 and about 1–2% median intervals. This fixture uses real registered-provider instance identities, but measures private state construction and shutdown, not a complete command. It does not establish a process-RSS change.

Ownership/security: new admission stops before the guard unlocks; each state mutex waits for its valid users. Cleanup does not hold the provider lock while files close. Concurrent instance close and repeated shutdown are covered. Borrowed streams remain open. Runtime shutdown can start before every instance is terminal; its own instance list remains alive through completion. The patch does not assume the opposite order.

### 5. Original Wasm input — Confirmed issue, fix rejected

The audit found public active-data slices into decoder input, including empty slices. Passive bytes and expression metadata are copied; decoded strings own their bytes; arbitrary custom payloads are not stored in public Compiled metadata. Private execution data is already a separate packed copy.

A 16-byte active segment with a 4 MiB unused custom section retained the input after all caller references were dropped. An empty active segment did too. A no-active-data control did not. The trial used one compact public backing allocation, separate from private execution bytes.

Twenty final pairs showed:

- Small-active-data/custom-section retained-heap delta: **4,119.80 → 15.81 KiB** after GC (p<0.001).
- Empty-active-data/custom-section retained-heap delta: **4,119.78 → 15.78 KiB** after GC (p<0.001).
- Data-heavy compile-plus-close: **397.4 → 447.7 µs, +12.68%** (p<0.001, median intervals about 2%). B/op rose **271.5 → 527.8 KiB**, +94.38%; allocations 54 → 56.
- Small compile: 9.334 → 9.580 µs, uncertain (p=0.104). Custom-section and no-data compile controls had no confirmed timing change.
- Normal custom-section heap endpoint stayed about 8.421 MiB. The post-GC improvement does not prove a normal RSS improvement.

The data-heavy slowdown exceeds the preselected 2% practical margin. The production copy was removed. `rejected-input` preserves the exact candidate and its benchmark/test source. Private execution isolation remains tested; public input retention is still an open issue. No extra compilation allocation is retained in the accepted work.

### 6. Native caches — Diagnosed, no production change

The engine cache holds one stack. Its requested size is bounded, and a different later stack-size request already closes the old mapping. The arena cache holds one mapping of at most 1 MiB; larger arenas already unmap on release.

The opt-in native test uses actual recursive Wasm execution and table instantiation, then closes each owner and runs a small instance. Mapping RSS/PSS, faults, heap and stack counters are logged without forced GC. The containing VMA can include adjacent owners; it is not added to overlapping process totals. Address reuse is identified separately from owner reuse.

Go 1.27 guarded observations: the 4 MiB default stack VMA kept about 2 MiB resident after the workload and small reuse. An 8 MiB stack kept about 4 MiB while idle; the later default stack request replaced its owner. A small table arena was reused; the oversized arena was unmapped. The revised test initializes all table entries. Its 320,304-byte arena is reused, while its 3,200,304-byte arena is released. The smaller arena shares a larger VMA, so the reported 1,760 KiB VMA RSS must not be called its own resident size. Stack RSS in this run consists of anonymous huge pages. The final log records 100 ms idle points and confirms owner replacement even when Linux reuses an address. These are diagnostic observations, not independent paired policy samples.

No cache policy experiment is accepted. Common-case and large-cache reuse latency, a twenty-pair policy comparison, and longer sustained-memory measurements were not completed. The observed idle storage is a bounded reuse cost, not proof of a leak. Guest linear-memory reclamation and protection were not changed.

### 7. Lazy filesystem state — Diagnosed, no production change

`stateFor` combines identity/lifetime validation with filesystem-state creation. Argument callbacks can create a fresh descriptor table for later instances. Provider start opens the initial preopens; later instances open their independent preopens at first state creation.

The new test uses two argument-only instances and confirms independent preopen owners. Renaming and replacing the host directory does not change the already-open directory identity. Simply moving creation to the first filesystem call would change that point in time. A safe split needs a separate identity/preopen owner, rollback and locking design. No callback bypass, deferred permission check or new state cache was added. Lazy first-use timing experiments were not completed.

### 8. Wider ownership sweep — Diagnosed, no production change

| Source/owner | Lifetime and growth | Finding/risk |
|---|---|---|
| Validator operand/control scratch, `wasm/validate.go:1076` | Reused within one module validation; module/function sizes affect growth | Not a process-global cache. Dropping it per function can restore repeated allocation. |
| Backend operand chunks, `amd64/stack.go:finishFunction` | Worker-owned; overflow above the existing retained-byte bound is dropped | Existing bounded release is present; no new scratch pool proposed. |
| Reference store, `reference_store.go:2790` | Reference and attachment lifetime | Terminal cleanup already drops maps and externref slices. Earlier release risks live handles. |
| Native GC collector, `gc/native/collector.go:222` | Collector lifetime; guest growth under resource policy | Close already drops scratch and owned heaps. Collection-cycle reuse is not terminal retention. |
| Compiled code, `code_mapping.go` | Compiled/instance owners | Explicit Close and corrected benchmark CodeImage ownership remain. Finalizers are fallback only. |
| Imports errors, `imports.go:61` | Imports lifetime; caller-controlled rejected operations | Duplicate snapshot validation is fixed. Repeated distinct invalid mutations are not changed. |
| Native engine and arena | One cached owner each, with existing size rules | Resident reuse cost remains; policy changes need latency evidence. |

This is a targeted source audit, not a proof that every long-lived structure has been checked.

## Validation and limits

The exact commands, exit status and sampled temperature for every build/test are in `checks/*.json`; output is in the matching logs.

| Check | Result |
|---|---|
| Go 1.27 guarded provider core and p1 race suites | Pass |
| Go 1.22 provider core and p1 normal suites | Pass |
| Wago import/caller/host race selection, after fixture setup | Pass |
| Wago new snapshot tests, execution-snapshot ownership and host-call allocation tests, Go 1.27 guarded and Go 1.22 normal | Pass |
| Terminal map and provider shutdown stress, including concurrent close | Pass |
| Path differential fuzz, 3 seconds, one worker | Pass |
| Four isolated AMD64 call controls, 10,000 calls each | 0 B/op and 0 allocs/op |
| Existing provider argument/environment allocation and isolation tests | Pass in the provider suites |
| Linux ARM64 guarded provider test binary and Wago production dependencies | Cross-build passes; not executed |

The first combined diagnostic/control process reported 1 B/op and rounded 0 allocs/op for add-one over 100 calls. That benchmark already warms the call; the byte cost was not attributed to a first call. It is preserved as an unexplained process-observation limit. The branch-hint benchmark was skipped because it is ARM64-only. The separate AMD64 arity controls then ran without the mapping diagnostic. Their timings are not used as performance comparisons, since the correctness wrapper throttles the process. The original 72-case execution set was not rerun; only these focused allocation controls were checked.

Expected baseline regression failures are kept. Two setup errors are also kept: a broad Wago check initially lacked the pinned conformance fixture, and a directly invoked test binary initially used the wrong working directory. Corrected runs passed. The early thermal stop is kept separately. TinyGo was not run; earlier TinyGo failures were not reclassified or worked around.
 `validation-index.json` distinguishes expected baseline failures, setup mistakes, and final checks. Native ARM64 execution was not available. The full 984-case suite was deliberately not run. No unrelated TinyGo fix or new TinyGo pass is claimed.

No accepted production change alters host dispatch, guest bounds, capabilities, compiler scheduling, guest-memory zeroing, or imported-memory ownership. The rejected copy experiment is absent from the final production diff. Every accepted patch has focused regression tests and a clean patch-application check.

Physical-memory limits remain explicit: timing benchmarks do different amounts of useful work in equal time; they are not fixed-work memory comparisons. The runner's fork-to-exit maximum can include inherited observer pages. Its 50 ms process sampling can miss short-lived processes. Those observations are saved in `process-memory-context.json`, but are not presented as per-instance peaks or normal idle-memory savings. Forced-GC heap checks prove reachability effects only. The overall physical-memory goal is **not complete**.

## Reproduction

Set `R=/tmp/wago-retention-roobe8pc` for the recorded worktrees. The captured `build.work` selects the local provider. For another location, create detached worktrees at the two revisions above and use the same workspace module selections. Apply the provider patches in order 01, 03, 04; apply Wago patch 02. The diagnostic patches are separate. `patch-apply-check.json` verifies this sequence against the recorded bases.

Build paired binaries with the same test source and `go test -c -tags wago_guardpage`. Each `*-overlay.json` selects the saved production baseline for its comparison; use the stage-specific candidate source below. Applying all accepted patches and omitting every candidate overlay would not reproduce the earlier stages. Source and binary hashes are in `*-build.json`. Candidate source hashes were checked against these saved copies:

| Case | Baseline production file | Candidate production file |
|---|---|---|
| path | `fs-path-baseline.go` | `fs-map-baseline.go` |
| fsmap | `fs-map-baseline.go` | `fs-shutdown-baseline.go` |
| shutdown | `fs-shutdown-baseline.go` | final provider `internal/core/fs.go` |
| snapshot | `imports-baseline.go` | final Wago `src/wago/imports.go` |
| input, rejected | `input-baseline.go` | `rejected-input/api.go.txt` |

Archived overlay sources have `.go.txt` suffixes under `overlay-inputs`; restore them as `.go` files outside either module. An overlay maps the absolute original production path to the selected saved source. Rebase the workspace and overlay paths when the work directory changes. Both builds must use the same diagnostic test source. The input experiment's exact original test source is archived as `rejected-input/input_retention_test.go.txt`. The rejected input comparison must use its preserved test source and candidate patch, not the accepted final tree.

Exact measurement invocations:

```sh
cd "$R/provider/internal/core"
python3 "$R/run-pairs.py" path
python3 "$R/run-pairs.py" fsmap
python3 "$R/run-pairs.py" shutdown
cd "$R/wago/src/wago"
python3 "$R/run-pairs.py" snapshot
python3 "$R/run-pairs.py" input
```

Each case file selects the exact leaf names and timing/count settings. Select a new output directory in the case file for a new run; the runner refuses an existing one. The final path evidence is `path-cool-pairs`, not the stopped `path-pairs` trial. Do not run these lines together with a test build.

```sh
python3 "$R/wago/docs/performance/setup-cleanup/continuation/summarize-wasi.py" \
  "$R/input-pairs" --output "$R/input-summary"
```

Use the same command for each run directory and a new summary destination. No accepted result requires a full-suite rerun.

JairusSW (approved):
looks good

